PR #16, opened 2023-09-22
Implement motion plan cache

methylDragon:
# WARNING
```
 *   This cache does NOT support collision detection!
 *   Plans will be put into and fetched from the cache IGNORING collision.
 *   If your planning scene is expected to change between cache lookups, do NOT
 *   use this cache, fetched plans are likely to result in collision then.
 *
 *   To handle collisions this class will need to hash the planning scene world
 *   msg (after zeroing out std_msgs/Header timestamps and sequences) and do an
 *   appropriate lookup. (very out of scope)
```

# Description

This PR adds a `MotionPlanCache` class that wraps a `warehouse_ros` database interface.
It is then incorporated into the motion_planning_server.

Features:
- Put and fetch `RobotTrajectory` plans keyed by `MotionPlanRequest` message values
  - This includes ALL joint, position, and orientation constraints!
- Supports fuzzy matching on the start and goal params in the request
- Supports overwriting "worse" plans with "better" plans (this prunes the database and keeps it tidy!)
- Separate caches for separate move groups

Furthermore, because our goals are stated in zero-offset TFs, the cache does a TF lookup and restates them in the database with respect to the workspace frame! (i.e. wrt. base_link.)

I could only test this on sqlite3. I couldn't set up MongoDB on my setup.

Also, while this is not 100% generalizable to all moveit plannable plans, I think this is a good candidate for upstreaming! :tada: 

# Why is this PR so big?
It's not trivial to key the motion plans. This is because there are many factors that go into the plan request, and the fulfilling plan. It should be visible from how I'm handling each factor on review.

I'll attach example requests, and an example generated trajectory. On inspecting those messages, it should be clear that it's not trivial to key them (but I think I managed to do it!)

## Start State (MotionPlanRequest) Variations

<details>
<summary>`is_diff=true`, which requires getting the current state</summary>

```
workspace_parameters:
  header:
    stamp:
      sec: 1695075603
      nanosec: 61941176
    frame_id: "base_link"
  min_corner:
    x: -0.570000
    y: -0.300000
    z: 0.00000
  max_corner:
    x: 0.830000
    y: 0.850000
    z: 1.39000
start_state:
  joint_state:
    header:
      stamp:
        sec: 0
        nanosec: 0
      frame_id: ""
    name: []
    position: []
    velocity: []
    effort: []
  multi_dof_joint_state:
    header:
      stamp:
        sec: 0
        nanosec: 0
      frame_id: ""
    joint_names: []
    transforms: []
    twist: []
    wrench: []
  attached_collision_objects: []
  is_diff: true
goal_constraints:
-
  name: ""
  joint_constraints: []
  position_constraints:
  -
    header:
      stamp:
        sec: 0
        nanosec: 0
      frame_id: "pickup_label_goal_pose"
    link_name: "gripper_link"
    target_point_offset:
      x: 0.00000
      y: 0.00000
      z: 0.00000
    constraint_region:  // (CH3) WE'RE IGNORING THIS
      primitives:
      -
        type: 2
        dimensions:
        - 0.000100000
        polygon:
          points: []
      primitive_poses:
      -
        position:
          x: 0.00000
          y: 0.00000
          z: -0.0100000
        orientation:
          x: 0.00000
          y: 0.00000
          z: 0.00000
          w: 1.00000
      meshes: []
      mesh_poses: []
    weight: 1.00000
  orientation_constraints:
  -
    header:
      stamp:
        sec: 0
        nanosec: 0
      frame_id: "pickup_label_goal_pose"
    orientation:
      x: 0.00000
      y: 0.00000
      z: 0.00000
      w: 1.00000
    link_name: "gripper_link"
    absolute_x_axis_tolerance: 0.000100000
    absolute_y_axis_tolerance: 0.000100000
    absolute_z_axis_tolerance: 0.000100000
    parameterization: 0
    weight: 1.00000
  visibility_constraints: []
path_constraints:
  name: ""
  joint_constraints: []
  position_constraints: []
  orientation_constraints: []
  visibility_constraints: []
trajectory_constraints:
  constraints: []
reference_trajectories: []
pipeline_id: ""
planner_id: ""
group_name: "manipulator"
num_planning_attempts: 1
allowed_planning_time: 5.00000
max_velocity_scaling_factor: 0.300000
max_acceleration_scaling_factor: 0.300000
cartesian_speed_end_effector_link: ""
max_cartesian_speed: 0.00000
```
</details>

<details>
<summary>Or if joint states were explicitly provided as start</summary>

```
workspace_parameters:
  header:
    stamp:
      sec: 1695079404
      nanosec: 4589339
    frame_id: "base_link"
  min_corner:
    x: -0.570000
    y: -0.300000
    z: 0.00000
  max_corner:
    x: 0.830000
    y: 0.850000
    z: 1.39000
start_state:
  joint_state:
    header:
      stamp:
        sec: 0
        nanosec: 0
      frame_id: "base_link"
    name:
    - "joint_1"
    - "joint_2"
    - "joint_3"
    - "joint_4"
    - "joint_5"
    - "joint_6"
    position:
    - 0.00000
    - 0.00000
    - 0.00000
    - 0.00000
    - 0.00000
    - 0.00000
    velocity: []
    effort: []
  multi_dof_joint_state:
    header:
      stamp:
        sec: 0
        nanosec: 0
      frame_id: "base_link"
    joint_names: []
    transforms: []
    twist: []
    wrench: []
  attached_collision_objects: []
  is_diff: false
goal_constraints:
-
  name: ""
  joint_constraints: []
  position_constraints:
  -
    header:
      stamp:
        sec: 0
        nanosec: 0
      frame_id: "pickup_label_goal_pose"
    link_name: "gripper_link"
    target_point_offset:
      x: 0.00000
      y: 0.00000
      z: 0.00000
    constraint_region:
      primitives:
      -
        type: 2
        dimensions:
        - 0.000100000
        polygon:
          points: []
      primitive_poses:
      -
        position:
          x: 0.00000
          y: 0.00000
          z: -0.0100000
        orientation:
          x: 0.00000
          y: 0.00000
          z: 0.00000
          w: 1.00000
      meshes: []
      mesh_poses: []
    weight: 1.00000
  orientation_constraints:
  -
    header:
      stamp:
        sec: 0
        nanosec: 0
      frame_id: "pickup_label_goal_pose"
    orientation:
      x: 0.00000
      y: 0.00000
      z: 0.00000
      w: 1.00000
    link_name: "gripper_link"
    absolute_x_axis_tolerance: 0.000100000
    absolute_y_axis_tolerance: 0.000100000
    absolute_z_axis_tolerance: 0.000100000
    parameterization: 0
    weight: 1.00000
  visibility_constraints: []
path_constraints:
  name: ""
  joint_constraints: []
  position_constraints: []
  orientation_constraints: []
  visibility_constraints: []
trajectory_constraints:
  constraints: []
reference_trajectories: []
pipeline_id: ""
planner_id: ""
group_name: "manipulator"
num_planning_attempts: 1
allowed_planning_time: 5.00000
max_velocity_scaling_factor: 0.300000
max_acceleration_scaling_factor: 0.300000
cartesian_speed_end_effector_link: ""
max_cartesian_speed: 0.00000
```
</details>

Notice for both cases, the goal is stated in terms of a TF frame (which can change.)
The motion plan cache restates the goal in terms of an x, y, z and orientation offset from the planning frame (which should not change.)

Planning frame is consequently part of the key for cache lookups.

## Generated Plan Example

<details>

```
joint_trajectory:
  header:
    stamp:
      sec: 0
      nanosec: 0
    frame_id: "base_link"
  joint_names:
  - "joint_1"
  - "joint_2"
  - "joint_3"
  - "joint_4"
  - "joint_5"
  - "joint_6"
  points:
  -
    positions:
    - 0.548301
    - 0.835924
    - 0.543629
    - -0.000849880
    - 0.191294
    - -1.02156
    velocities:
    - 0.0100000
    - 0.000238418
    - 0.00142367
    - -0.00630237
    - -0.00170240
    - -0.00574253
    accelerations:
    - 10.0000
    - 0.238418
    - 1.42367
    - -6.30237
    - -1.70240
    - -5.74253
    effort: []
    time_from_start:
      sec: 0
      nanosec: 0
  -
    positions:
    - 0.598300
    - 0.837116
    - 0.550747
    - -0.0323615
    - 0.182782
    - -1.05027
    velocities:
    - 1.00681
    - 0.0240042
    - 0.143337
    - -0.634529
    - -0.171400
    - -0.578163
    accelerations:
    - 10.0000
    - 0.238418
    - 1.42367
    - -6.30237
    - -1.70240
    - -5.74253
    effort: []
    time_from_start:
      sec: 0
      nanosec: 100000000
  -
    positions:
    - 0.748297
    - 0.840692
    - 0.572102
    - -0.126895
    - 0.157246
    - -1.13641
    velocities:
    - 2.00296
    - 0.0477543
    - 0.285157
    - -1.26234
    - -0.340985
    - -1.15021
    accelerations:
    - 10.0000
    - 0.238418
    - 1.42367
    - -6.30237
    - -1.70240
    - -5.74253
    effort: []
    time_from_start:
      sec: 0
      nanosec: 200000000
  -
    positions:
    - 0.987439
    - 0.846394
    - 0.606148
    - -0.277611
    - 0.116535
    - -1.27373
    velocities:
    - 2.34000
    - 0.0557899
    - 0.333140
    - -1.47476
    - -0.398362
    - -1.34375
    accelerations:
    - -10.0000
    - -0.238418
    - -1.42367
    - 6.30237
    - 1.70240
    - 5.74253
    effort: []
    time_from_start:
      sec: 0
      nanosec: 300000000
  -
    positions:
    - 1.17155
    - 0.850783
    - 0.632359
    - -0.393642
    - 0.0851921
    - -1.37946
    velocities:
    - 1.34000
    - 0.0319480
    - 0.190772
    - -0.844518
    - -0.228122
    - -0.769498
    accelerations:
    - -10.0000
    - -0.238418
    - -1.42367
    - 6.30237
    - 1.70240
    - 5.74253
    effort: []
    time_from_start:
      sec: 0
      nanosec: 400000000
  -
    positions:
    - 1.25565
    - 0.852788
    - 0.644333
    - -0.446650
    - 0.0708737
    - -1.42776
    velocities:
    - 0.340000
    - 0.00810622
    - 0.0484049
    - -0.214281
    - -0.0578817
    - -0.195246
    accelerations:
    - -10.0000
    - -0.238418
    - -1.42367
    - 6.30237
    - 1.70240
    - 5.74253
    effort: []
    time_from_start:
      sec: 0
      nanosec: 500000000
  -
    positions:
    - 1.26147
    - 0.852927
    - 0.645161
    - -0.450315
    - 0.0698835
    - -1.43110
    velocities:
    - 3.00625e-16
    - 7.16745e-18
    - 4.27992e-17
    - -1.89465e-16
    - -5.11785e-17
    - -1.72635e-16
    accelerations:
    - -10.0000
    - -0.238418
    - -1.42367
    - 6.30237
    - 1.70240
    - 5.74253
    effort: []
    time_from_start:
      sec: 0
      nanosec: 534110678
multi_dof_joint_trajectory:
  header:
    stamp:
      sec: 0
      nanosec: 0
    frame_id: ""
  joint_names: []
  points: []
```

</details>

## Insight
Notice that keying them requires considering many things, like the varying number of joints, joint constraints; orientation constraints; position constraints; planning frames, etc.

I picked what I think is a good set that covers most of our cases. What I am explicitly not supporting is stated below.

# Images

## Cache hits and pushes!
image

## A view of the schema
image

## With database collections for each move group
image

## Populated Entries
image

# Notes
Does not support:
- Constraint regions
- Collision objects
- Cartesian planning (introduced in #18)
- Velocity/Acceleration constraints
- Obstacles in the planning scene

It also only partially handles matching in cases where constraints get jumbled up in order. (But this really shouldn't happen, I think.)

# TODO
- [x] Retrofit the launch files (but what path do I use for the db?)
- [ ] Test on physical setup. I was only able to test by spoofing "execute_trajectory=True" and not actually executing the trajectory.
  - The fuzzy matching parameters should be tuned depending on how much precision we want. (I currently set them to about ~0.5 degrees per joint for the start, and 1mm goal offsets in all axes.)
- [ ] Unit tests??
- [ ] Benchmark cache time vs planning times?

### DIFF
--- a/nexus_motion_planner/src/motion_plan_cache.cpp
+++ b/nexus_motion_planner/src/motion_plan_cache.cpp
@@ -1,0 +1,918 @@
+#include <memory>
+#include <vector>
+
+#include "moveit/robot_state/conversions.h"
+#include "moveit/robot_state/robot_state.h"
+#include "warehouse_ros/database_loader.h"
+
+#include "rclcpp/rclcpp.hpp"
+#include "rclcpp/logging.hpp"
+#include "motion_plan_cache.hpp"
+
+namespace nexus {
+namespace motion_planner {
+
+#define NEXUS_MATCH_RANGE(arg, tolerance) \
+  arg - tolerance / 2, arg + tolerance / 2
+
+using warehouse_ros::MessageWithMetadata;
+using warehouse_ros::Metadata;
+using warehouse_ros::Query;
+
+MotionPlanCache::MotionPlanCache(const rclcpp::Node::SharedPtr& node)
+: node_(node)
+{
+  // TODO: methyldragon -
+  //   Declare or pass in these in the motion_planner_server instead?
+  if (!node_->has_parameter("warehouse_plugin"))
+  {
+    node_->declare_parameter<std::string>(
+      "warehouse_plugin", "warehouse_ros_sqlite::DatabaseConnection");
+  }
+
+  tf_buffer_ =
+    std::make_unique<tf2_ros::Buffer>(node_->get_clock());
+  tf_listener_ =
+    std::make_shared<tf2_ros::TransformListener>(*tf_buffer_);
+
+  warehouse_ros::DatabaseLoader loader(node_);
+  db_ = loader.loadDatabase();
+}
+
+void MotionPlanCache::init(
+  const std::string& db_path, uint32_t db_port, double exact_match_precision)
+{
+  RCLCPP_INFO(
+    node_->get_logger(),
+    "Opening motion plan cache database at: %s (Port: %d, Precision: %f)",
+    db_path.c_str(), db_port, exact_match_precision);
+
+  exact_match_precision_ = exact_match_precision;
+  db_->setParams(db_path, db_port);
+  db_->connect();
+}
+
+// TOP LEVEL OPS ===============================================================
+std::vector<MessageWithMetadata<moveit_msgs::msg::RobotTrajectory>::ConstPtr>
+MotionPlanCache::fetchAllMatchingPlans(
+  const moveit::planning_interface::MoveGroupInterface& move_group,
+  const std::string& move_group_namespace,
+  const moveit_msgs::msg::MotionPlanRequest& plan_request,
+  double start_tolerance, double goal_tolerance, bool metadata_only)
+{
+  auto coll = db_->openCollection<moveit_msgs::msg::RobotTrajectory>(
+    "move_group_plan_cache", move_group_namespace);
+
+  Query::Ptr query = coll.createQuery();
+
+  bool start_ok = this->extractAndAppendStartToQuery(
+    *query, move_group, plan_request, start_tolerance);
+  bool goal_ok = this->extractAndAppendGoalToQuery(
+    *query, move_group, plan_request, goal_tolerance);
+
+  if (!start_ok || !goal_ok)
+  {
+    RCLCPP_ERROR(node_->get_logger(), "Could not construct plan query.");
+    return {};
+  }
+
+  return coll.queryList(
+    query, metadata_only, /* sort_by */ "execution_time_s", true);
+}
+
+MessageWithMetadata<moveit_msgs::msg::RobotTrajectory>::ConstPtr
+MotionPlanCache::fetchBestMatchingPlan(
+  const moveit::planning_interface::MoveGroupInterface& move_group,
+  const std::string& move_group_namespace,
+  const moveit_msgs::msg::MotionPlanRequest& plan_request,
+  double start_tolerance, double goal_tolerance, bool metadata_only)
+{
+  // First find all matching, but metadata only.
+  // Then use the ID metadata of the best plan to pull the actual message.
+  auto matching_plans = this->fetchAllMatchingPlans(
+    move_group, move_group_namespace,
+    plan_request, start_tolerance, goal_tolerance,
+    true);
+
+  if (matching_plans.empty())
+  {
+    RCLCPP_INFO(node_->get_logger(), "No matching plans found.");
+    return nullptr;
+  }
+
+  auto coll = db_->openCollection<moveit_msgs::msg::RobotTrajectory>(
+    "move_group_plan_cache", move_group_namespace);
+
+  // Best plan is at first index, since the lookup query was sorted by
+  // execution_time.
+  int best_plan_id = matching_plans.at(0)->lookupInt("id");
+  Query::Ptr best_query = coll.createQuery();
+  best_query->append("id", best_plan_id);
+
+  return coll.findOne(best_query, metadata_only);
+}
+
+bool
+MotionPlanCache::putPlan(
+  const moveit::planning_interface::MoveGroupInterface& move_group,
+  const std::string& move_group_namespace,
+  const moveit_msgs::msg::MotionPlanRequest& plan_request,
+  const moveit_msgs::msg::RobotTrajectory& plan,
+  double execution_time_s,
+  bool overwrite)
+{
+  // Check pre-conditions
+  if (!plan.multi_dof_joint_trajectory.points.empty())
+  {
+    RCLCPP_ERROR(
+      node_->get_logger(),
+      "Skipping insert: Multi-DOF trajectory plans are not supported.");
+    return false;
+  }
+  if (plan_request.workspace_parameters.header.frame_id.empty() ||
+    plan.joint_trajectory.header.frame_id.empty())
+  {
+    RCLCPP_ERROR(
+      node_->get_logger(), "Skipping insert: Frame IDs cannot be empty.");
+    return false;
+  }
+  if (plan_request.workspace_parameters.header.frame_id !=
+    plan.joint_trajectory.header.frame_id)
+  {
+    RCLCPP_ERROR(
+      node_->get_logger(),
+      "Skipping insert: "
+      "Plan request frame (%s) does not match plan frame (%s).",
+      plan_request.workspace_parameters.header.frame_id.c_str(),
+      plan.joint_trajectory.header.frame_id.c_str());
+    return false;
+  }
+
+  auto coll = db_->openCollection<moveit_msgs::msg::RobotTrajectory>(
+    "move_group_plan_cache", move_group_namespace);
+
+  // If start and goal are "exact" match, AND the candidate plan is better,
+  // overwrite.
+  Query::Ptr exact_query = coll.createQuery();
+
+  bool start_query_ok = this->extractAndAppendStartToQuery(
+    *exact_query, move_group,
+    plan_request, exact_match_precision_);
+  bool goal_query_ok = this->extractAndAppendGoalToQuery(
+    *exact_query, move_group,
+    plan_request, exact_match_precision_);
+
+  if (!start_query_ok || !goal_query_ok)
+  {
+    RCLCPP_ERROR(
+      node_->get_logger(),
+      "Skipping insert: Could not construct overwrite query.");
+    return false;
+  }
+
+  auto exact_matches = coll.queryList(exact_query, /* metadata_only */ true);
+  double best_execution_time_seen = std::numeric_limits<double>::infinity();
+  if (!exact_matches.empty())
+  {
+    for (auto& match : exact_matches)
+    {
+      double match_execution_time_s =
+        match->lookupDouble("execution_time_s");
+      if (match_execution_time_s < best_execution_time_seen)
+      {
+        best_execution_time_seen = match_execution_time_s;
+      }
+
+      if (match_execution_time_s > execution_time_s)
+      {
+        // If we found any "exact" matches that are worse, delete them.
+        if (overwrite)
+        {
+          int delete_id = match->lookupInt("id");
+          RCLCPP_INFO(
+            node_->get_logger(),
+            "Overwriting plan (id: %d): "
+            "execution_time (%es) > new plan's execution_time (%es)",
+            delete_id, match_execution_time_s, execution_time_s);
+
+          Query::Ptr delete_query = coll.createQuery();
+          delete_query->append("id", delete_id);
+          coll.removeMessages(delete_query);
+        }
+      }
+    }
+  }
+
+  // Insert if candidate is best seen.
+  if (execution_time_s < best_execution_time_seen)
+  {
+    Metadata::Ptr insert_metadata = coll.createMetadata();
+
+    bool start_meta_ok = this->extractAndAppendStartToMetadata(
+      *insert_metadata, move_group, plan_request);
+    bool goal_meta_ok = this->extractAndAppendGoalToMetadata(
+      *insert_metadata, move_group, plan_request);
+    insert_metadata->append("execution_time_s", execution_time_s);
+
+    if (!start_meta_ok || !goal_meta_ok)
+    {
+      RCLCPP_ERROR(
+        node_->get_logger(),
+        "Skipping insert: Could not construct insert metadata.");
+      return false;
+    }
+
+    RCLCPP_INFO(
+      node_->get_logger(),
+      "Inserting plan: New plan execution_time (%es) "
+      "is better than best plan's execution_time (%es)",
+      execution_time_s, best_execution_time_seen);
+
+    coll.insert(plan, insert_metadata);
+    return true;
+  }
+
+  RCLCPP_INFO(
+    node_->get_logger(),
+    "Skipping insert: New plan execution_time (%es) "
+    "is worse than best plan's execution_time (%es)",
+    execution_time_s, best_execution_time_seen);
+  return false;
+}
+
+// QUERY CONSTRUCTION ==========================================================
+bool
+MotionPlanCache::extractAndAppendStartToQuery(
+  Query& query,
+  const moveit::planning_interface::MoveGroupInterface& move_group,
+  const moveit_msgs::msg::MotionPlanRequest& plan_request,
+  double match_tolerance)
+{
+  match_tolerance += exact_match_precision_;
+
+  // Make ignored members explicit
+  if (!plan_request.start_state.multi_dof_joint_state.joint_names.empty())
+  {
+    RCLCPP_WARN(
+      node_->get_logger(),
+      "Ignoring start_state.multi_dof_joint_states: Not supported.");
+  }
+  if (!plan_request.start_state.attached_collision_objects.empty())
+  {
+    RCLCPP_WARN(
+      node_->get_logger(),
+      "Ignoring start_state.attached_collision_objects: Not supported.");
+  }
+
+  // auto original = *query;  // Copy not supported.
+
+  query.append("group_name", plan_request.group_name);
+
+  // Workspace params
+  // Match anything within our specified workspace limits.
+  query.append(
+    "workspace_parameters.header.frame_id",
+    plan_request.workspace_parameters.header.frame_id);
+  query.appendGTE(
+    "workspace_parameters.min_corner.x",
+    plan_request.workspace_parameters.min_corner.x);
+  query.appendGTE(
+    "workspace_parameters.min_corner.y",
+    plan_request.workspace_parameters.min_corner.y);
+  query.appendGTE(
+    "workspace_parameters.min_corner.z",
+    plan_request.workspace_parameters.min_corner.z);
+  query.appendLTE(
+    "workspace_parameters.max_corner.x",
+    plan_request.workspace_parameters.max_corner.x);
+  query.appendLTE(
+    "workspace_parameters.max_corner.y",
+    plan_request.workspace_parameters.max_corner.y);
+  query.appendLTE(
+    "workspace_parameters.max_corner.z",
+    plan_request.workspace_parameters.max_corner.z);
+
+  // Joint state
+  //   Only accounts for joint_state position. Ignores velocity and effort.
+  if (plan_request.start_state.is_diff)
+  {
+    // If plan request states that the start_state is_diff, then we need to get
+    // the current state from the move_group.
+
+    // NOTE: methyldragon -
+    //   I think if is_diff is on, the joint states will not be populated in all
+    //   of our motion plan requests? If this isn't the case we might need to
+    //   apply the joint states after as well.
+    moveit::core::RobotStatePtr current_state = move_group.getCurrentState();
+    if (!current_state)
+    {
+      RCLCPP_WARN(
+        node_->get_logger(),
+        "Skipping start query append: Could not get robot state.");
+      // *query = original;  // Undo our changes. (Can't. Copy not supported.)
+      return false;
+    }
+
+    moveit_msgs::msg::RobotState current_state_msg;
+    robotStateToRobotStateMsg(*current_state, current_state_msg);
+
+    for (size_t i = 0; i < current_state_msg.joint_state.name.size(); i++)
+    {
+      query.append(
+        "start_state.joint_state.name_" + std::to_string(i),
+        current_state_msg.joint_state.name.at(i));
+      query.appendRangeInclusive(
+        "start_state.joint_state.position_" + std::to_string(i),
+        NEXUS_MATCH_RANGE(
+          current_state_msg.joint_state.position.at(i), match_tolerance)
+      );
+    }
+  }
+  else
+  {
+    for (
+      size_t i = 0; i < plan_request.start_state.joint_state.name.size(); i++
+    )
+    {
+      query.append(
+        "start_state.joint_state.name_" + std::to_string(i),
+        plan_request.start_state.joint_state.name.at(i));
+      query.appendRangeInclusive(
+        "start_state.joint_state.position_" + std::to_string(i),
+        NEXUS_MATCH_RANGE(
+          plan_request.start_state.joint_state.position.at(i), match_tolerance)
+      );
+    }
+  }
+  return true;
+}
+
+bool
+MotionPlanCache::extractAndAppendGoalToQuery(
+  Query& query,
+  const moveit::planning_interface::MoveGroupInterface& /* move_group */,
+  const moveit_msgs::msg::MotionPlanRequest& plan_request,
+  double match_tolerance)
+{
+  match_tolerance += exact_match_precision_;
+
+  // Make ignored members explicit
+  bool emit_position_constraint_warning = false;
+  for (auto& constraint : plan_request.goal_constraints)
+  {
+    for (auto& position_constraint : constraint.position_constraints)
+    {
+      if (!position_constraint.constraint_region.primitives.empty())
+      {
+        emit_position_constraint_warning = true;
+        break;
+      }
+    }
+    if (emit_position_constraint_warning)
+    {
+      break;
+    }
+  }
+  if (emit_position_constraint_warning)
+  {
+    RCLCPP_WARN(
+      node_->get_logger(),
+      "Ignoring goal_constraints.position_constraints.constraint_region: "
+      "Not supported.");
+  }
+
+  // auto original = *query;  // Copy not supported.
+
+  query.appendRangeInclusive(
+    "max_velocity_scaling_factor",
+    NEXUS_MATCH_RANGE(
+      plan_request.max_velocity_scaling_factor, match_tolerance)
+  );
+  query.appendRangeInclusive(
+    "max_acceleration_scaling_factor",
+    NEXUS_MATCH_RANGE(
+      plan_request.max_acceleration_scaling_factor, match_tolerance)
+  );
+  query.appendRangeInclusive(
+    "max_cartesian_speed",
+    NEXUS_MATCH_RANGE(plan_request.max_cartesian_speed, match_tolerance));
+
+  // Extract constraints (so we don't have cardinality on goal_constraint idx.)
+  std::vector<moveit_msgs::msg::JointConstraint> joint_constraints;
+  std::vector<moveit_msgs::msg::PositionConstraint> position_constraints;
+  std::vector<moveit_msgs::msg::OrientationConstraint> orientation_constraints;
+  for (auto& constraint : plan_request.goal_constraints)
+  {
+    for (auto& joint_constraint : constraint.joint_constraints)
+    {
+      joint_constraints.push_back(joint_constraint);
+    }
+    for (auto& position_constraint : constraint.position_constraints)
+    {
+      position_constraints.push_back(position_constraint);
+    }
+    for (auto& orientation_constraint : constraint.orientation_constraints)
+    {
+      orientation_constraints.push_back(orientation_constraint);
+    }
+  }
+
+  // Joint constraints
+  size_t joint_idx = 0;
+  for (auto& constraint : joint_constraints)
+  {
+    std::string meta_name =
+      "goal_constraints.joint_constraints_" + std::to_string(joint_idx++);
+
+    query.append(meta_name + ".joint_name", constraint.joint_name);
+    query.appendRangeInclusive(
+      meta_name + ".position",
+      NEXUS_MATCH_RANGE(constraint.position, match_tolerance));
+    query.appendGTE(
+      meta_name + ".tolerance_above", constraint.tolerance_above);
+    query.appendLTE(
+      meta_name + ".tolerance_below", constraint.tolerance_below);
+  }
+
+  // Position constraints
+  // All offsets will be "frozen" and computed wrt. the workspace frame
+  // instead.
+  if (!position_constraints.empty())
+  {
+    query.append(
+      "goal_constraints.position_constraints.header.frame_id",
+      plan_request.workspace_parameters.header.frame_id);
+
+    size_t pos_idx = 0;
+    for (auto& constraint : position_constraints)
+    {
+      std::string meta_name =
+        "goal_constraints.position_constraints_" + std::to_string(pos_idx++);
+
+      // Compute offsets wrt. to workspace frame.
+      double x_offset = 0;
+      double y_offset = 0;
+      double z_offset = 0;
+
+      if (plan_request.workspace_parameters.header.frame_id !=
+        constraint.header.frame_id)
+      {
+        try
+        {
+          auto transform = tf_buffer_->lookupTransform(
+            constraint.header.frame_id,
+            plan_request.workspace_parameters.header.frame_id,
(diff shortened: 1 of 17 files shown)


nexus_motion_planner/src/motion_plan_cache.cpp
```diff
@@ -0,0 +1,920 @@
+#include <memory>
```
Yadunund, line 1: Add copyright header

nexus_motion_planner/src/motion_planner_server.hpp
```diff
@@ -88,6 +95,19 @@ class MotionPlannerServer : public rclcpp_lifecycle::LifecycleNode
   double _workspace_max_y;
   double _workspace_max_z;
 
+  // Motion plan caching
+  std::shared_ptr<nexus::motion_planner::MotionPlanCache> _motion_plan_cache;
```
Yadunund, line 99:
```suggestion
  std::unique_ptr<nexus::motion_planner::MotionPlanCache> _motion_plan_cache;
```

nexus_motion_planner/src/motion_planner_server.hpp
```diff
@@ -88,6 +95,19 @@ class MotionPlannerServer : public rclcpp_lifecycle::LifecycleNode
...
+
+  bool _use_motion_plan_cache;
+  bool _use_cached_plans_instead_of_planning;
+  bool _overwrite_worse_plans;
+  std::string _cache_db_plugin;
+  std::string _cache_db_host;
+  int _cache_db_port;
+  double _cache_exact_match_tolerance;
```
Yadunund, line 107: Instead of three doubles, let's define an enum class for each of these cases. Then have two params: 1) To get the string name of the enum type and match as defined above 2) The value of the tolerance. You could also consider using `std::variant` here and define 3 three different structs with boolean values of tolerance.

nexus_motion_planner/src/motion_planner_server.hpp
```diff
@@ -88,6 +95,19 @@ class MotionPlannerServer : public rclcpp_lifecycle::LifecycleNode
   double _workspace_max_y;
   double _workspace_max_z;
 
+  // Motion plan caching
+  std::shared_ptr<nexus::motion_planner::MotionPlanCache> _motion_plan_cache;
+
+  bool _use_motion_plan_cache;
```
Yadunund, line 101: Instead of three bools, let's define class which defines how the databse will be used for better extensibility.
```
enum class PlannerDatabaseMode: public unint8_t
{
  // The planner will generate a plan. Find if an entry exists in the databse and overwrite if execution time is better.
  Training,
  
  // Try to lookup a cached plan and if non is available, plan live but do not update the cache.
  Optional,

  // Return a plan only if present in the database cache.
  // ReadOnly, 
};

```

And below we would store an optional of this type.
```
std::optional<PlannerDatabaseMode> = std::nullopt;
```

We can parse a similarly named ROS param which takes in a string with the same name as the enum variants above. If we have a matching name, we set the member to the appropriate enum; Else leave it as optional. In the service response, if this member is nullopt, do a live plan (same as now). Else if not nullopt, plan and update database accordingly.
We should also instantiate the `_motion_plan_cache` only if this variable is not nullopt and other database specific parameters are defined (port, etc).

nexus_motion_planner/src/motion_plan_cache.cpp
```diff
@@ -0,0 +1,920 @@
...
+    node_->declare_parameter<std::string>(
+      "warehouse_plugin", "warehouse_ros_sqlite::DatabaseConnection");
+  }
+
+  tf_buffer_ =
+    std::make_unique<tf2_ros::Buffer>(node_->get_clock());
+  tf_listener_ =
+    std::make_shared<tf2_ros::TransformListener>(*tf_buffer_);
+
+  warehouse_ros::DatabaseLoader loader(node_);
```
Yadunund, line 38: I don't think we should use the `DatabaseLoader` class as it is forcing us to store (and spin) a separate ROS node in the MotionPlanner class. Looking at it's [implementation](https://github.com/ros-planning/warehouse_ros/blob/ros2/src/database_loader.cpp), it is a simple wrapper to retrieve ROS Params and load the right storage plugin. We're already retrieving some of the params in the `MotionPlanner` class, we can also try loading the plugin within the `on_configure()` and directly instantiate the `db` object via pluginlib. This would save us the need to spin another node which brings the total node count to three! (we already have one for this server, one for the move_group_interface).

Yadunund (changes requested):
Thanks for great effort here! Did a first pass review with some comments.

I'm concerned we're not using the `APIs from PlanningSceneStorage` class that is already implemented upstream for such caching. It looks like they [serialize the MotionPlanRequest msg](https://github.com/ros-planning/moveit2/blob/c9a9e407b0a27858c8eea838ccbbd389d7109f9e/moveit_ros/warehouse/src/planning_scene_storage.cpp#L111-L113) to check if it's cached before updating the database. Is there a reason we're not able to simply reuse the approach there? 

Lastly, as part of this implementation we should also
* Update the nexus_msgs/stv/GetMotionPlan request to add a boolean to force lookup from the database (ReadOnly in the review comments) and fail the response if the trajectory does not exist in the databse.
* Update the PlanMotion BT node to add a similar input port as above which will append the value to the request.

methylDragon:
> Thanks for great effort here! Did a first pass review with some comments.
> 
> I'm concerned we're not using the `APIs from PlanningSceneStorage` class that is already implemented upstream for such caching. It looks like they [serialize the MotionPlanRequest msg](https://github.com/ros-planning/moveit2/blob/c9a9e407b0a27858c8eea838ccbbd389d7109f9e/moveit_ros/warehouse/src/planning_scene_storage.cpp#L111-L113) to check if it's cached before updating the database. Is there a reason we're not able to simply reuse the approach there?

Hey Yadu, I just saw this review! Thanks for the review!

The `moveit_warehouse` classes don't support fuzzy lookups. That class is more used for saving and loading scenes and states in the RViz window (as per [the tutorial](https://ros-planning.github.io/moveit_tutorials/doc/persistent_scenes_and_states/persistent_scenes_and_states.html)), and so only needs the ability to label messages with user defined IDs for saving and retrieval. (It then does an exact match on these IDs when looking the entries up to load them.)

Whereas for us, we need to be able to do fuzzy lookups to match "close enough" start and goal states to support caching and lookups for the purposes of planning.

> Lastly, as part of this implementation we should also
> 
> * Update the nexus_msgs/stv/GetMotionPlan request to add a boolean to force lookup from the database (ReadOnly in the review comments) and fail the response if the trajectory does not exist in the databse.
> * Update the PlanMotion BT node to add a similar input port as above which will append the value to the request.

Ah, I had put it as a parameter on motion planning service's node, but I can just as easily put it in the motion plan request instead.
Will do!

nexus_motion_planner/src/motion_plan_cache.hpp
```diff
@@ -0,0 +1,138 @@
...
+    const moveit_msgs::msg::MotionPlanRequest& plan_request);
+
+private:
+  rclcpp::Node::SharedPtr node_;
+  warehouse_ros::DatabaseConnection::Ptr db_;
+
+  double exact_match_precision_ = 0;
+
+  std::unique_ptr<tf2_ros::Buffer> tf_buffer_;
+  std::shared_ptr<tf2_ros::TransformListener> tf_listener_;
```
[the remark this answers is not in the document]
methylDragon: The tf is just to pull the state of the robot at time of planning (to then explicitly fill into the cache when adding the cache entry.) (It is also to pull the state of the goal pose wrt. the planning frame (`base_link`) and explicitly save it.

This is because there are certain motion plan requests that call for the current state of the robot (without actually telling the planner what the current state of the robot is, so it has to look it up.) (Same with the goal.)

See the `is_diff=true`, which requires getting the current state` variant of motion plan request.

In all of these cases (where plan request entries are "dynamic", the caching class I implement explicitly looks up the values to statically instantiate them as joint states or poses relative to a common, unchanging frame (`base_link`, or whatever was defined as the workspace frame.)

At no point in time are cache entries edited after they are inserted (unless they are deleted because a better entry was found.)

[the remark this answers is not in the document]
methylDragon: e.g.

A goal constraint:
```
  position_constraints:
  -
    header:
      stamp:
        sec: 0
        nanosec: 0
      frame_id: "pickup_label_goal_pose"
    link_name: "gripper_link"
    target_point_offset:
      x: 0.00000
      y: 0.00000
      z: 0.00000
```

Will be converted at plan time into:
```
```
  position_constraints:
  -
    header:
      stamp:
        sec: 0
        nanosec: 0
      frame_id: "base_link"
    link_name: "gripper_link"
    target_point_offset:
      x: SOME_NUM
      y: SOME_NUM
      z: SOME_NUM
```

Where the pose is the transform from `base_link` to `pickup_label_goal_pose` at time of motion plan request.

File: nexus_motion_planner/src/motion_plan_cache.hpp [hunk 7]
```diff
@@ -0,0 +1,138 @@
...
+    double start_tolerance, double goal_tolerance, bool metadata_only = false);
+
+  bool put_plan(
+    const moveit::planning_interface::MoveGroupInterface& move_group,
+    const std::string& move_group_namespace,
+    const moveit_msgs::msg::MotionPlanRequest& plan_request,
+    const moveit_msgs::msg::RobotTrajectory& plan,
+    double execution_time_s,
+    double planning_time_s,
+    bool overwrite = true);
```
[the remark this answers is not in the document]
methylDragon: In most cases, no. However:
- If `exact_match_precision` is set very high (relative to start and goal match tolerance), then overwriting will result in the cache being pretty sparse.
- Not overwriting results in the cache space being maximally dense, which maximizes the chances of a cache hit on lookup.
- Also it can be used to see historically how plans have evolved over time.

I think for all those reasons I'd like to keep a parameter here to disable overwriting, especially if we plan to upstream this cache class to MoveIt!

Note: Overwriting in the context of the cache means deleting **ALL** "worse" plans that have an "exact" match (according to the `exact_match_precision` amount, across all cache entry metadata fields), before inserting the new "better" entry.

nexus_motion_planner/src/motion_plan_cache.hpp
```diff
@@ -0,0 +1,138 @@
...
+  bool extract_and_append_goal_to_metadata(
+    warehouse_ros::Metadata& metadata,
+    const moveit::planning_interface::MoveGroupInterface& move_group,
+    const moveit_msgs::msg::MotionPlanRequest& plan_request);
+
+private:
+  rclcpp::Node::SharedPtr node_;
+  warehouse_ros::DatabaseConnection::Ptr db_;
+
+  double exact_match_precision_ = 0;
```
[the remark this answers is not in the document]
methylDragon: Ah, I set this as a default on the ROS param, but no harm setting it on the variable as well!

(Note: I noticed that, even on sim when the robot is commanded to the same location each time, a value of 0.001 (~0.05 degrees per joint) seemed more appropriate.)

nexus_motion_planner/src/motion_planner_server.hpp
methylDragon: Any cache will use all three of these tolerance values (exact matches for population, and start and goal for plan fetches.)

More importantly, these values should be configurable by the user.

I don't understand why we would use an enum or variant in that case?
Maybe I'm missing something here?

nexus_motion_planner/src/motion_plan_cache.hpp
```diff
@@ -0,0 +1,138 @@
+#ifndef SRC__MOTION_PLAN_CACHE_HPP
```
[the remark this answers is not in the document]
methylDragon: https://github.com/osrf/nexus/pull/16/commits/49edb376ed5b2b80b21a37c183d01e6fb9fa01c8

nexus_motion_planner/src/motion_plan_cache.hpp
```diff
@@ -0,0 +1,138 @@
...
+#include <geometry_msgs/msg/pose_stamped.hpp>
+#include <lifecycle_msgs/msg/state.hpp>
+
+// moveit modules
+#include <moveit/move_group_interface/move_group_interface.h>
+
+// NEXUS messages
+#include <nexus_endpoints.hpp>
+
+namespace nexus {
```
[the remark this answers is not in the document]
methylDragon: https://github.com/osrf/nexus/pull/16/commits/49edb376ed5b2b80b21a37c183d01e6fb9fa01c8

nexus_motion_planner/src/motion_plan_cache.cpp
methylDragon: https://github.com/osrf/nexus/pull/16/commits/49edb376ed5b2b80b21a37c183d01e6fb9fa01c8

nexus_motion_planner/src/motion_planner_server.hpp
methylDragon: e605c7f12747b7ed1ebbc0ca91f37e65381c1382

nexus_motion_planner/src/motion_plan_cache.hpp
```diff
@@ -0,0 +1,138 @@
...
+ * For example, the "my_move_group" move group will have its cache stored in
+ * `move_group_plan_cache@my_move_group`
+ *
+ * Motion plans are keyed on:
+ *   - Plan Start: robot joint state
+ *   - Plan Goal (either of):
+ *     - Final pose (wrt. `planning_frame` (usually `base_link`))
+ *     - Final robot joint states
+ *
+ * Motion plans may be looked up with some tolerance.
```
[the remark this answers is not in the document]
methylDragon: There is no ROS param that sets the tolerance here.
Instead it's passed in per lookup call, since this class is meant to be a standalone class.

nexus_motion_planner/src/motion_plan_cache.hpp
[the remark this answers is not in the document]
methylDragon: cd0d55daec3b151b8374f2f3c0cd87f6472dbfd4

nexus_motion_planner/src/motion_plan_cache.cpp
methylDragon: We still need a node for the tf listener..
I will just use the motion planner's node for both the tf listener and the database loading.

nexus_motion_planner/src/motion_planner_server.hpp
methylDragon: d85907d118b82ffd8ab1efb798f8b71d9c66fc78

nexus_motion_planner/src/motion_plan_cache.cpp
koonpeng, line 1: I am working on remove exceptions throughout the codebase. Most of `warehouse_ros` functions may throw exceptions, can you add a `TODO` at the top of the file to remind to catch all these exceptions?

nexus_motion_planner/src/motion_plan_cache.hpp
```diff
@@ -0,0 +1,165 @@
...
+  fetch_all_matching_plans(
+    const moveit::planning_interface::MoveGroupInterface& move_group,
+    const std::string& move_group_namespace,
+    const moveit_msgs::msg::MotionPlanRequest& plan_request,
+    double start_tolerance, double goal_tolerance, bool metadata_only = false);
+
+  warehouse_ros::MessageWithMetadata<
+    moveit_msgs::msg::RobotTrajectory
+  >::ConstPtr
+  fetch_best_matching_plan(
```
koonpeng, line 113: Can you add docs explaining how the best plan is determined.

nexus_motion_planner/src/motion_plan_cache.cpp
```diff
@@ -0,0 +1,935 @@
...
+  if (plan_request.start_state.is_diff)
+  {
+    // If plan request states that the start_state is_diff, then we need to get
+    // the current state from the move_group.
+
+    // NOTE: methyldragon -
+    //   I think if is_diff is on, the joint states will not be populated in all
+    //   of our motion plan requests? If this isn't the case we might need to
+    //   apply the joint states as offsets as well.
+    moveit::core::RobotStatePtr current_state = move_group.getCurrentState();
```
koonpeng, line 323: iiuc this can be flimsy, the state we retrieve here may not be the same as the state moveit used for planning. How much impact would it be if we drop support for `is_diff`?

nexus_motion_planner/src/motion_plan_cache.cpp
```diff
@@ -0,0 +1,935 @@
...
+    }
+    for (auto& position_constraint : constraint.position_constraints)
+    {
+      position_constraints.push_back(position_constraint);
+    }
+    for (auto& orientation_constraint : constraint.orientation_constraints)
+    {
+      orientation_constraints.push_back(orientation_constraint);
+    }
+  }
```
koonpeng, line 435: These vectors doesn't seem to be modified after these insertions, can't we avoid the copy and using `plan_request.goal_constraints` directly? (will require double loop though).

nexus_motion_planner/src/motion_planner_server.hpp
koonpeng: I think enums and variants don't quite make sense here also. Perhaps there is confusion with the naming having both "exact_match" and "tolerance", the fomer suggest a bool and the latter suggest a double.

nexus_motion_planner/src/motion_plan_cache.hpp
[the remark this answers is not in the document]
koonpeng: I think this behavior should be documented, iirc `put_plan` only inserts a new plan if it is better than existing plans. So `overwrite` has no effect if it is not the best plan.

koonpeng (changes requested):
There is quite a lot of logic going on in maintaining the cache, the lack of tests is abit worrying. But I will leave it to @Yadunund to decide if this is a blocker.

methylDragon:
I'll address all the comments in a separate PR (the branching structure makes it difficult to be updating it here, since changes need to be duplicated across the two PRs at the moment.)

nexus_motion_planner/src/motion_plan_cache.cpp
```diff
@@ -0,0 +1,935 @@
...
+
+#include "rclcpp/rclcpp.hpp"
+#include "rclcpp/logging.hpp"
+#include "motion_plan_cache.hpp"
+
+namespace nexus {
+namespace motion_planner {
+
+#define NEXUS_MATCH_RANGE(arg, tolerance) \
+  arg - tolerance / 2, arg + tolerance / 2
```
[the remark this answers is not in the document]
methylDragon: 6ebd330b82fd970d64424f1e2d3b3d8d50af56ee

nexus_motion_planner/src/motion_plan_cache.cpp
```diff
@@ -0,0 +1,935 @@
...
+using warehouse_ros::Query;
+
+MotionPlanCache::MotionPlanCache(const rclcpp::Node::SharedPtr& node)
+: node_(node)
+{
+  if (!node_->has_parameter("warehouse_plugin"))
+  {
+    node_->declare_parameter<std::string>(
+      "warehouse_plugin", "warehouse_ros_sqlite::DatabaseConnection");
+  }
```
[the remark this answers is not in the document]
methylDragon: If it exists with a different db provider, it only works if the entry is valid (corresponds to a plugin that exists in the workspace.)

We won't know ahead of time what that list is, and warehouse_ros will throw the error for us if it does.

I added a comment to make this clearer.

76f5faf057be46461ebbcc1f6ad12d273908f816 in https://github.com/osrf/nexus/pull/26

nexus_motion_planner/src/motion_plan_cache.cpp
```diff
@@ -0,0 +1,935 @@
...
+  const std::string& db_path, uint32_t db_port, double exact_match_precision)
+{
+  RCLCPP_INFO(
+    node_->get_logger(),
+    "Opening motion plan cache database at: %s (Port: %d, Precision: %f)",
+    db_path.c_str(), db_port, exact_match_precision);
+
+  exact_match_precision_ = exact_match_precision;
+  db_->setParams(db_path, db_port);
+  db_->connect();
```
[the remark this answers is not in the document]
methylDragon: 4e76d7192f18b3bcfd8f5bd234e7768002b724b8 in https://github.com/osrf/nexus/pull/26

nexus_motion_planner/src/motion_planner_server.cpp
```diff
@@ -205,6 +327,13 @@ auto MotionPlannerServer::on_configure(const LifecycleState& /*state*/)
 -> CallbackReturn
 {
   RCLCPP_INFO(this->get_logger(), "Configuring...");
+
+  if (_cache_mode != PlannerDatabaseMode::Unset)
+  {
+    _motion_plan_cache->init(
+      _cache_db_host, _cache_db_port, _cache_exact_match_tolerance);
```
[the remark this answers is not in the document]
methylDragon: 03afb8e898fa76ef41507727368769927de1d0ae

nexus_motion_planner/src/motion_plan_cache.cpp
methylDragon: 0e84dd4576d070a1ed11d1cdac9d492ca5aedaf7 in https://github.com/osrf/nexus/pull/26

nexus_motion_planner/src/motion_plan_cache.hpp
methylDragon: Added comment, and even made it configurable now (defaults to `execution_time_s`, no behavior change.

6004f96fc39d4b287ab5c142adb1fc220c83a4b8 in https://github.com/osrf/nexus/pull/26

nexus_motion_planner/src/motion_plan_cache.cpp
methylDragon: `is_diff` is the default way plans are made (if drawn from a start state from a TF frame.)

The `extract_and_append_plan_start_to_query` function is used for fetching plans, so we get the current state of the robot to pull up plans in the cache which are close-enough to the current state of the robot, wrt. **concrete** joint-state values.

This way, we avoid that flimsiness you mentioned.

nexus_motion_planner/src/motion_plan_cache.cpp
methylDragon: warehouse_ros doesn't support array/vector columns.

What is happening here is we are inserting the constraints as separate columns.

E.g.:
- joint_constraint_1
- joint_constraint_2
- joint_constraint_3
- position_constraint_1
- position_constraint_2
- position_constraint_3
- orientation_constraint_1
- orientation_constraint_2
- orientation_constraint_3

If we used `goal_constraints` directly, like so, we would have:
- constraint_1
- constraint_2
- constraint_3
- constraint_4
- constraint_5
- constraint_6

The problem with doing this is it makes it much more likely that constraint ordering will matter, when we don't actually care about it when pulling up our plans.

E.g. for goal constraints that are ordered differently but still have the same constraint:
- [joint_1, position_1, joint_2, position_2]
- [joint_1, joint_2, position_1, position_2]

Then the inserted columns will have different ordering if we didn't extract out the constraint types, which means lookups will be unnecessarily sensitive to ordering (between the different types of constraints.)

nexus_motion_planner/src/motion_planner_server.hpp
methylDragon: I'll add comments to explain:
ccbf91cd9f98e626bd1ba4a535b5b2aff10eba0c in https://github.com/osrf/nexus/pull/26

nexus_motion_planner/src/motion_plan_cache.hpp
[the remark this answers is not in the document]
methylDragon: The insertion check only happens at the end, so technically speaking `overwrite` will delete all worse plans.

I decided to name it `overwrite` -> `delete_worse_plans` to be more clear.

f5713a1620c41a1f78adc6ffc899cac89f6bd64a in https://github.com/osrf/nexus/pull/26

nexus_motion_planner/src/motion_plan_cache.cpp
koonpeng: My concern is like so:

1. Motion plan server receives plan request with `is_diff`.
2. Pass request to moveit, moveit get the current robot state and does the planning.
3. Motion plan server receive the plan, get the current robot state again and use it to key the cache.
4. At step 2, we may get joint 1 = `0.5`, then at step 3, we get joint 1 = `0.501`. This can cause "cache poisoning" if the difference is too big.

It would be better if there is a way to get the exact state moveit used to do the planning, or is that infact what `move_group.getCurrentState()` is doing? I'm not experienced with moveit so I thought it is getting the robot state from `ros_control`.

nexus_motion_planner/src/motion_plan_cache.cpp
koonpeng: iiuc, we should also sort the constraints? e.g.
```
goal_constraints[0] = some joint_2 constraint
goal_constraints[1] = some joint_1 constraint
```
should be stored with stable ordering (`joint_1`, `joint_2`).

methylDragon: Fair point!
f166c890ea598f3bbba6a1db4f0ed957352a589c

nexus_motion_planner/src/motion_plan_cache.cpp
methylDragon: Hmm, I think getCurrentState() is getting it from the planning scene monitor, so most likely ros_control (unless the state was set manually otherwise)...

I don't think this is a problem we can solve at this level, and since we want to upstream this (so maintenance is up to the MoveIt maintainers), I don't think it's a good idea to drop support for `is_diff` too...

Ideally when it gets upstreamed, the caching should happen as part of the plan call within MoveIt itself.

I don't think this is a problem in our application, so I think we should TODO this for now.

bbc07cec581f5b127ec7f86227da69b078994a38

koonpeng (approved):
Pre-approving for when the tests are merged.

Yadunund (approved):
Thanks for this amazing effort @methylDragon!

PR #18, opened 2023-10-18
Implement cartesian plan cache

methylDragon:
# WARNING
```
 *   This cache does NOT support collision detection!
 *   Plans will be put into and fetched from the cache IGNORING collision.
 *   If your planning scene is expected to change between cache lookups, do NOT
 *   use this cache, fetched plans are likely to result in collision then.
 *
 *   To handle collisions this class will need to hash the planning scene world
 *   msg (after zeroing out std_msgs/Header timestamps and sequences) and do an
 *   appropriate lookup. (very out of scope)
```

# Description
This PR builds on top of:
- https://github.com/osrf/nexus/pull/16

It implements and uses the plan cache for cartesian plans!

Cached plans appear in the `move_group_cartesian_plan_cache` collection in the db:
image

### DIFF
--- a/nexus_capabilities/src/capabilities/plan_motion.cpp
+++ b/nexus_capabilities/src/capabilities/plan_motion.cpp
@@ -120,6 +120,17 @@ make_request()
   req->max_velocity_scaling_factor = scale_speed;
   req->max_acceleration_scaling_factor = scale_speed;
 
+  auto maybe_read_only =
+    this->getInput<bool>("force_cache_mode_execute_read_only");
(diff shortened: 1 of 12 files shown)


koonpeng (changes requested):
Same comments as #16 wrt `warehouse_ros` exceptions and lack of tests.

methylDragon:
Will implement changes in a separate PR.

Yadunund (approved):
`force_cache_mode_execute_read_only` feels like a mouthful but we can iterate on this in the future. Thanks for this effort!

PR #21, opened 2023-11-06
Add Changelogs and bump version to 0.1.0

Yadunund:
This PR adds Changelogs to all the pkgs here and bumps the versions to `0.1.0` (`0.0.1` is reserved for `humble` version of the pkgs).

Once this PR is merged, I will tag the latest commit as `0.1.0`.

### DIFF
--- a/nexus_capabilities/CHANGELOG.rst
+++ b/nexus_capabilities/CHANGELOG.rst
@@ -2,6 +2,6 @@
 Changelog for package nexus_capabilities
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Forthcoming
------------
+0.1.0 (2023-11-06)
+------------------
 * Provides abstract interfaces for defining new capabilities along with several out-of-the-box capabilities.
--- a/nexus_capabilities/CHANGELOG.rst
+++ b/nexus_capabilities/CHANGELOG.rst
@@ -1,0 +1,7 @@
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Changelog for package nexus_capabilities
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Forthcoming
+-----------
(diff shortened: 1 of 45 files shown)


nexus_capabilities/CHANGELOG.rst
koonpeng, line 1: Any reason why rst is used instead of md?

koonpeng, line 1: Will the changelogs be auto generated? If so, we need a guideline on how to write commit messages. If not, we need a guideline on how to write the changelogs.

Yadunund: [REP 132](https://www.ros.org/reps/rep-0132.html) specifies that Changelogs should be rst. It is the format we use for all ROS 2 and Open-RMF pkgs.

Yadunund: These changelogs are generated using ros-infrastructure tooling such `catkin_generate_changelogs` and `catkin_prepare_release` (the latter will automatically bump pkg versions and update the changelog with the newly tagged version.
In the future when we run `catkin_generate_changelogs`, it will automatically populate the CHANGELOG.rst file with any new commits since the last tag and use the commit message as the description.

koonpeng: Is there a REP on writing commit messages?

> with any new commits since the last tag and use the commit message as the description

What if I have commits that does not deserve a change log? e.g. Updating github actions, cleanups etc.

koonpeng: Fair enough, though imo this process can be improved. Most change log management tools analyzes commit messages to determine if they should be included in the changelogs, and what category is the change (bug fixes, new features). I thought the ROS tools would have commit message formats to support those things as well.

koonpeng (approved):

nexus_capabilities/CHANGELOG.rst
Yadunund: Ah that would be a valuable addition to REP 132 which can motivate changes to `catkin_generate_changelogs`.

PR #23, opened 2023-11-09
WIP remove exceptions

koonpeng:
This is a preview of the work in removing exceptions.

Inspired from rust, `Result<T>` is the alternative to exceptions. Internally it uses `std::variant` to store either the result or a `std::exception`.

* `[[nodiscard]]` should almost always be used when returning `Result<T>`. We don't have custom lint rules so it will have to be checked manually.
* `Result<void>` is a special case that using `std::optional` instead.
* One downside is that we cannot return a reference, `Result<T&>` does not work because `std::variant` cannot contain references.
  * In theory should be able to specialize `Result<T>` for all reference types and store a pointer internally instead.


Yadunund:
Could you merge the latest main into the base branch also rebase/update this branch?

luca-della-vedova:
I'm wondering, especially now that we bumped main to 24.04 and all the compiler versions are fairly recent, would it make sense to increase our C++ standard to C++23 and use [std::expected](https://en.cppreference.com/w/cpp/utility/expected) rather than manually implementing our own `Result` type?

koonpeng:
> I'm wondering, especially now that we bumped main to 24.04 and all the compiler versions are fairly recent, would it make sense to increase our C++ standard to C++23 and use [std::expected](https://en.cppreference.com/w/cpp/utility/expected) rather than manually implementing our own `Result` type?

I'm down for that

PR #24, opened 2023-11-10
Pass same NodeOptions to planner's internal node

Yadunund:
The `nexus_motion_planner` spins an internal `rclcpp::Node::SharedPtr` since `moveit's` `move_group_interface` API does is not compatible with a `LifecycleNode`.  Looking at the implementation, we were trying to declare some of the move_group related parameters in the LifecycleNode and then "forward" the declaration to the internal node. However there is no need to do this since we can just pass the same `NodeOptions` to the internal node which will also pass the `arguments` to the ROS 2 node which the move_group_interface will rely on to declare relevant parameters. 

Without this fix, I was seeing a couple of warning in the terminal from not declaring certain essential moveit parameters as arguments were missing. But now by passing the same options, the internal node has access to the same arguments as the parent and can declare all required moveit params.

Also fix linking error with running the test executable.

### DIFF
--- a/nexus_motion_planner/src/motion_planner_server.cpp
+++ b/nexus_motion_planner/src/motion_planner_server.cpp
@@ -20,17 +20,14 @@ MotionPlannerServer::MotionPlannerServer(const rclcpp::NodeOptions& options)
 {
   RCLCPP_INFO(this->get_logger(), "Motion Planner Server is running...");
 
-  auto internal_node_options = rclcpp::NodeOptions();
-  internal_node_options.automatically_declare_parameters_from_overrides(true);
-  internal_node_options.use_global_arguments(false);
   _internal_node = std::make_shared<rclcpp::Node>(
-    "motion_planner_server_internal_node", internal_node_options);
+    "motion_planner_server_internal_node", options);
   _spin_thread = std::thread(
-    [this]()
+    [node = _internal_node]()
     {
       while (rclcpp::ok())
       {
-        rclcpp::spin_some(_internal_node);
+        rclcpp::spin_some(node);
       }
     });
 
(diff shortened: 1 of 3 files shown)


nexus_motion_planner/src/motion_planner_server.cpp
```diff
@@ -20,17 +20,14 @@ MotionPlannerServer::MotionPlannerServer(const rclcpp::NodeOptions& options)
...
-    "motion_planner_server_internal_node", internal_node_options);
+    "motion_planner_server_internal_node", options);
   _spin_thread = std::thread(
-    [this]()
+    [node = _internal_node]()
     {
       while (rclcpp::ok())
       {
-        rclcpp::spin_some(_internal_node);
+        rclcpp::spin_some(node);
```
koonpeng, line 30: Not a blocker, I just find this problematic. Not just the fact that we are spinning a node within another, but also that it is using the default executor, not sure about the api of executors but spinning an already spinning executor seems like something problematic. Likely we never ran into any issues because we are using component nodes, which iirc does not use the default executor.

koonpeng (approved):

nexus_motion_planner/src/motion_planner_server.cpp
Yadunund: Some clarifications
* We're doing this because the move_group_interface API requires an `rclcpp::Node::SharedPtr` for instantiation.
* `rclcpp::spin_some(node)` will create a new executor and spin that executor and not spin the same default executor. See https://github.com/ros2/rclcpp/blob/411dbe8212dd753ee60871f8605d0a8beaae5e37/rclcpp/src/rclcpp/executors.cpp#L32-L37
* Although this is a component, we have always been running it as a standalone node.

koonpeng: Interesting that `rclcpp::spin_some` creates a new executor each time, wouldn't that cause some heavy overhead?

Yadunund: It probably does. We should switch to `rclcpp::spin()`

PR #28, opened 2023-11-22
Add unit tests for motion plan cache

methylDragon:
# Description
Built on top of https://github.com/osrf/nexus/pull/26 (and all its parents)

Adds comprehensive tests for motion and cartesian plan caching.
The testing uncovered some small bugs which are also fixed in this PR.

Run tests using:
```
clear && colcon test --packages-select nexus_motion_planner --event-handlers console_direct+
```

# Notes
I apologize in advance for the way I'm doing the testing. Since we need move_group to run, we need launch tests. I tried to make it as neat as possible.

Also I figured using robot descriptions in MoveIt would be best.

Not much way around it if we want easy assertions to bubble up if tests fail.


nexus_motion_planner/CMakeLists.txt
```diff
@@ -72,6 +77,8 @@ install(
 )
 
 if(BUILD_TESTING)
+  find_package(ament_cmake_gtest REQUIRED)
```
koonpeng, line 80: Doesn't seem to be used?

nexus_motion_planner/src/test_motion_plan_cache.cpp
```diff
@@ -0,0 +1,1164 @@
...
+    move_group->setStartState(*curr_state);
+
+    test_motion_plans(move_group, cache);
+    test_cartesian_plans(move_group, cache);
+  }
+
+  rclcpp::shutdown();
+  spin_thread.join();
+
+  return 0;
```
koonpeng, line 1158: Return non-zero if test fail.

nexus_motion_planner/src/test_motion_plan_cache.cpp
```diff
@@ -0,0 +1,1164 @@
...
+void check_and_emit(
+  bool predicate, const std::string& prefix, const std::string& label)
+{
+  if (predicate)
+  {
+    std::cout << "[PASS] " << prefix << ": " << label << std::endl;
+  }
+  else
+  {
+    std::cout << "[FAIL] " << prefix << ": " << label << std::endl;
```
koonpeng, line 40: Consider exiting the program if there is a failure. Or have 2 helper functions, one to continue on failure and another to exit immediately.

koonpeng (changes requested):

koonpeng:
Still looking at the test contents.

nexus_motion_planner/src/test_motion_plan_cache.cpp
methylDragon: https://github.com/osrf/nexus/pull/28/commits/a06c57815925a4911231b86974f896b6e03928db

nexus_motion_planner/src/test_motion_plan_cache.cpp
methylDragon: https://github.com/osrf/nexus/pull/28/commits/a06c57815925a4911231b86974f896b6e03928db

nexus_motion_planner/CMakeLists.txt
methylDragon: 3e2bb014750b8a350cb747804d6f43fae188d814

nexus_motion_planner/src/motion_plan_cache.hpp
```diff
@@ -81,7 +78,7 @@ namespace motion_planner {
  *
  * Motion plans may be looked up with some tolerance at call time.
  */
-class MotionPlanCache
+class MotionPlanCache : public std::enable_shared_from_this<MotionPlanCache>
```
koonpeng, line 81: Since this has `enable_shared_from_this`, I'm assuming that it must be created as a `shared_ptr`. If so, make the constructor private and have a public static "make" function to create an instance. This prevents accidentally creating a non `shared_ptr` instance which result in undefined behaviors.

---

On second look, I don't see any `*_from_this` used? Do we need this inheritance?

nexus_motion_planner/src/test_motion_plan_cache.cpp
```diff
@@ -0,0 +1,1165 @@
...
+  planning_time = 999;
+
+  check_and_emit(
+    cache->put_plan(
+      *move_group, g_robot_name, plan_req, plan,
+      execution_time, planning_time, false),
+    prefix, "Put first valid plan, no delete_worse_plans, ok");
+
+  check_and_emit(
+    cache->count_plans(g_robot_name) == 1, prefix, "One plan in cache");
```
koonpeng, line 267: nit: Add a check for different robot name, ensuring that it returns no result. Might be overkill to do this check everytime we check the cache, so I think checking when the cache is "most populated" is good enough.

nexus_motion_planner/src/test_motion_plan_cache.cpp
```diff
@@ -0,0 +1,1165 @@
...
+  fetched_plan = cache->fetch_best_matching_plan(
+    *move_group, g_robot_name, close_matching_plan_req, 0, 0);
+
+  check_and_emit(fetched_plans.size() == 0, prefix, "Fetch all returns empty");
+  check_and_emit(
+    fetched_plan == nullptr, prefix, "Fetch best plan is nullptr");
+
+  // Fetch non-matching, in tolerance
+  //
+  // Close key within tolerance limit should have cache hit
```
koonpeng, line 347: Add test where only one of start or goal tolerance is within tolerance.

nexus_motion_planner/test/test_motion_plan_cache.py
```diff
@@ -0,0 +1,398 @@
...
+    ]
+
+    for prefix, labels in test_cases:
+        for label in labels:
+            process_tools.assert_output_sync(
+                launch_context,
+                motion_cache_test_runner_node,
+                validate_stream(f"[PASS] {prefix}: {label}"),
+                timeout=10
+            )
```
koonpeng, line 386: Unless I'm missing something, I don't think we need to scan the output, the main test exits with non-zero if anything fails, all we need is to catch that and report the result accordingly.

imo this is also not good design, it is repeating the test case and labels, making modification unnecessarily errior-prone and tedious. It may also break if the uut code prints something to stdout.

koonpeng (changes requested):

nexus_motion_planner/src/motion_plan_cache.hpp
methylDragon: I had a brainfart and put it in when I had to make a shared_ptr of the class.

Fixed.
bf53bf7a4906794ff48fd6a2d05ecd39d4191fa3

nexus_motion_planner/src/test_motion_plan_cache.cpp
methylDragon: a235d207d1a90b0e13a6ef7d2032568437971266

nexus_motion_planner/src/test_motion_plan_cache.cpp
methylDragon: 776c5fdbefa8443ca9a73b87c17ae22d24d710de

PR #29, opened 2023-11-23
Add force_cache_mode_execute_read_only

methylDragon:
Builds on top of: https://github.com/osrf/nexus/pull/28

Adds the ability to force a GetMotionPlan service request to use motion cache mode ExecuteReadOnly regardless of the server's cache mode, on a per request basis.

### DIFF
--- a/nexus_capabilities/src/capabilities/plan_motion.cpp
+++ b/nexus_capabilities/src/capabilities/plan_motion.cpp
@@ -120,6 +120,17 @@ make_request()
   req->max_velocity_scaling_factor = scale_speed;
   req->max_acceleration_scaling_factor = scale_speed;
 
+  auto maybe_read_only =
+    this->getInput<bool>("force_cache_mode_execute_read_only");
+  if (maybe_read_only.has_value())
+  {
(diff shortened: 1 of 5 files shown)


nexus_capabilities/src/capabilities/plan_motion.cpp
```diff
@@ -120,6 +120,17 @@ make_request()
   req->max_velocity_scaling_factor = scale_speed;
   req->max_acceleration_scaling_factor = scale_speed;
 
+  auto maybe_read_only =
+    this->getInput<bool>("force_cache_mode_execute_read_only");
```
koonpeng, line 124: The port should be defined in `providedPorts` (in the header) as well.

koonpeng (changes requested):

nexus_capabilities/src/capabilities/plan_motion.cpp
methylDragon: Oops: d8f4cda74c1b854f48e579ad8f06355cc3f88525

koonpeng (approved):

PR #30, opened 2023-12-05
Fix and add motion planner server tests

methylDragon:
Builds on top of:
- https://github.com/osrf/nexus/pull/16

I uncommented the old launch test, and discovered it was passing.

Then I tried adding a test for the motion plan cache, and had it fail because we need access to the robot joint state.

Turns out the old tests weren't loading a ros2_control controller, so I fixed it (maybe that's what broke in the first place?)

After building, test with:
```
clear && colcon test --packages-select nexus_motion_planner --event-handlers=console_direct+
```

### DIFF
--- a/.github/workflows/nexus_integration_tests.yaml
+++ b/.github/workflows/nexus_integration_tests.yaml
@@ -40,6 +40,8 @@ jobs:
         rosdep update
         rosdep install --from-paths . -yir
     - name: build
-      run: /ros_entrypoint.sh colcon build --packages-up-to nexus_integration_tests --mixin release lld --cmake-args -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
-    - name: test
+      run: /ros_entrypoint.sh colcon build --packages-up-to nexus_integration_tests nexus_motion_planner --mixin release lld --cmake-args -DCMAKE_CXX_COMPILER_LAUNCHER=ccache
+    - name: Test - Unit Tests
+      run: . ./install/setup.bash && RMW_IMPLEMENTATION=rmw_cyclonedds_cpp /ros_entrypoint.sh colcon test --packages-select nexus_motion_planner --event-handlers=console_direct+
+    - name: Test - Integration Test
       run: . ./install/setup.bash && cd nexus_integration_tests && RMW_IMPLEMENTATION=rmw_cyclonedds_cpp /ros_entrypoint.sh python3 -m unittest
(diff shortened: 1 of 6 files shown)


nexus_motion_planner/test/test_request.test.py
```diff
@@ -179,6 +187,34 @@ def generate_test_description():
...
 
+    # Load ros2_control controller (so joint states are published)
+    # This needs to be loaded in tandem with the MoveIt controller manager
+    ros2_controllers_path = os.path.join(
+        get_package_share_directory(moveit_config_package),
+        "config",
+        "ros2_controllers.yaml",
+    )
+    ros2_control_node = Node(
+        package="controller_manager",
```
Yadunund, line 198: Add `<test_depend>` for ros2_control pkgs.

nexus_motion_planner/test/test_request.test.py
```diff
@@ -318,6 +355,8 @@ def generate_test_description():
         trigger_lifecycle_5,
         trigger_lifecycle_6,
         trigger_get_plan_irb1300_poses,
+        ros2_control_node,
+        *load_controllers,
```
Yadunund, line 359: What does the `*` do? Is it a pythonic way to append to a list?

Yadunund (changes requested):
Looks good. Minor comments. Should we also update the integration test CI job to run `colcon test` for this pkg?

nexus_motion_planner/test/test_request.test.py
methylDragon: The `*` unpacks the list into its composite elements (you can think of it as an `extend()` but in the middle of the list.

It was introduced to avoid "clunky syntax handling". https://peps.python.org/pep-3132/ , so I think it's acceptable.

The technically simpler way to do it would be:
```

    return launch.LaunchDescription([
        robot_state_publisher,
        motion_planner,
        move_group_node,
        static_tf_node_1,
        static_tf_node_2,
        trigger_lifecycle_1,
        trigger_lifecycle_2,
        trigger_lifecycle_3,
        trigger_lifecycle_4,
        trigger_lifecycle_5,
        trigger_lifecycle_6,
        trigger_get_plan_irb1300_poses,
        ros2_control_node]
  + load_controllers
  + [launch_testing.actions.ReadyToTest()
    ]), node_mapping
```

But I didn't like the triple addition.

nexus_motion_planner/test/test_request.test.py
methylDragon: 51f34f4c6170d6cc978e2be34190e8efee8833f8

methylDragon:
I added the tests to CI https://github.com/osrf/nexus/pull/30/commits/b18e5b353b5d79c44396a9de6b52146b4722bba4

Yadunund (approved):

PR #31, opened 2024-01-22
Return error if not plan was fetched for non-cartesian requests

Yadunund:
Follow up to #16.

Enforce plan to be fetched if request containts `force_cache_mode_execute_read_only=true`. This logic already exists for the cartesian requests.


methylDragon (approved):
Ah, good catch!

PR #35, opened 2024-10-18
Migrate to Jazzy

luca-della-vedova:
This PR migrates the Nexus codebase to Ubuntu 24.04 and ROS 2 Jazzy. Overview of the changes below, I tried to keep the commits split, clean and self descriptive to ease the review.

### Zenoh

I bumped the zenoh bridge to `1.0.0`, this required a minor migration but integration tests seem to still be OK.

### Style

Moving to Ubuntu 24.04 brings a new uncrustify version that would result in a lot of C++ file changes. I just removed uncrustify from CI for now and will create a clean PR with it.

### Lifecycle nodes

During migration I found an upstream issue in `rclcpp` in how timers are cleaned up that broke the workcell orchestrator. Added a workaround and a TODO in https://github.com/osrf/nexus/pull/35/commits/1f0f43a086056bf0483b004c39c79e43f3917277

### Commented test

I managed to get all the tests to pass except one that however was failing already so I didn't focus on trying to get it right. I added a note in https://github.com/osrf/nexus/pull/35/commits/b683b1f82425be612be6cd7dd2f81a3221f198c1

## Test it!

All the tests other than the one that I noted above pass, including integration tests. I also did a few manual runs and it seems to be OK. Note however that this has not been tested on hardware.


luca-della-vedova:
I gave a try to using clang and there are compile errors due to ros2-control https://github.com/ros-controls/realtime_tools/pull/183

.github/workflows/nexus_integration_tests.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
       run: |
         apt update && apt install -y git curl libclang-dev
     - name: Setup Rust for nexus_zenoh_bridge
-      uses: dtolnay/rust-toolchain@stable
+      uses: dtolnay/rust-toolchain@1.75
```
Yadunund, line 23: We can remove this setup step entirely and instead rely on `rosdep` to install `cargo` as system dep version of `cargo` matches what we need.

.github/workflows/nexus_workcell_editor.yaml
```diff
@@ -26,13 +26,17 @@ jobs:
...
+        pip install colcon-ros-cargo --break-system-packages
     - uses: actions/checkout@v2
     - name: vcs
+      # TODO(luca) Go back to cloning a tag when a new version is released with jazzy repos file
       run: |
-        git clone https://github.com/ros2-rust/ros2_rust.git -b 0.4.0
-        vcs import . < ros2_rust/ros2_rust_iron.repos
+        git clone https://github.com/ros2-rust/ros2_rust.git
+        cd ros2_rust
+        git checkout f45a66f47dc727e3ccb13037a6c57923af1446c7
```
Yadunund, line 37: Can we checkout a semantic tag instead of a commit hash?

abb.repos
```diff
@@ -17,5 +17,5 @@ repositories:
     version: rolling
   thirdparty/abb/abb_ros2:
     type: git
-    url: https://github.com/yadunund/abb_ros2.git
-    version: main
+    url: https://github.com/luca-della-vedova/abb_ros2.git
```
Yadunund, line 20: This can be reverted now that the PR from your fork was merged.

nexus_common/src/action_client_bt_node_test.cpp
```diff
@@ -16,7 +16,7 @@
  */
 
 #define CATCH_CONFIG_MAIN
-#include <catch2/catch.hpp>
+#include <rmf_utils/catch.hpp>
```
Yadunund, line 19: What is the rationale for this change?

Also, please upgrade `rmf_utils` to `<build_depend>` https://github.com/osrf/nexus/blob/728fd0d366f9bc65da9c318d64a0c5e8c385cd89/nexus_common/package.xml#L25

nexus_zenoh_bridge_dds_vendor/CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ find_package(ament_cmake_vendor_package REQUIRED)
 
 ament_vendor(zeno_bridge_dds_vendor
   VCS_URL https://github.com/eclipse-zenoh/zenoh-plugin-dds.git
-  VCS_VERSION 0.7.2-rc
+  VCS_VERSION 1.0.0
```
Yadunund, line 15: Use the latest patch?

```suggestion
  VCS_VERSION 1.0.2
```

nexus_workcell_orchestrator/src/workcell_orchestrator.cpp
```diff
@@ -895,7 +895,8 @@ void WorkcellOrchestrator::_register()
       RCLCPP_INFO(this->get_logger(),
         "Successfully registered with system orchestrator");
       this->_register_timer->cancel();
-      this->_register_timer.reset();
+      // TODO(luca) reintroduce once https://github.com/ros2/rclcpp/issues/2652 is fixed and released
+      // this->_register_timer.reset();
```
Yadunund, line 899: What is the implication of not resetting the timer? More CPU consumption? Can we reset it outside the service callback later on?

Should we consider switching to the events executor as suggested https://github.com/ros2/rclcpp/issues/2652#issuecomment-2419198033?

Yadunund (changes requested):
Awesome work with this migration! Left a little bit of feedback to consider.

.github/workflows/nexus_workcell_editor.yaml
luca-della-vedova: Nope, the [last release](https://github.com/ros2-rust/ros2_rust/releases/tag/0.4.1) was in November last year, the commit that introduced the jazzy repos file was [three weeks ago](https://github.com/ros2-rust/ros2_rust/pull/425)

nexus_workcell_orchestrator/src/workcell_orchestrator.cpp
luca-della-vedova: I find the `reset` name a bit confusing in this context, since we have a pointer to a timer it can be easily misunderstood:

```
timer->reset(); // Resets the timer, but it still exists
timer.reset(); // Resets the pointer, effectively destroying the timer
```

Since we cancel the timer it will not run anymore so I believe the overhead is really just the fact that we still have memory allocated for it. I would definitely not recommend changing the whole executor for the overhead of a cancelled timer being allocated that will supposedly be fixed soon, but given how efficient the event executor seems I would consider experimenting with it and see if it helps in a future PR.

.github/workflows/nexus_integration_tests.yaml
luca-della-vedova: https://github.com/osrf/nexus/pull/35/commits/6cc59f376de6d1f13d303559366b726d4a80f68a

abb.repos
luca-della-vedova: https://github.com/osrf/nexus/pull/35/commits/3d1a51dc8b3d789a296d56741e76df60ac6e6abb

nexus_common/src/action_client_bt_node_test.cpp
luca-della-vedova: I haven't actually looked into it but I did now! Reason, compilation fails without:

```
--- stderr: nexus_common                                                                                                                                                                                
[...]/src/nexus/nexus_common/src/action_client_bt_node_test.cpp:19:10: fatal error: catch2/catch.hpp: No such file or directory
   19 | #include <catch2/catch.hpp>
      |          ^~~~~~~~~~~~~~~~~~
compilation terminated.
```

Explanation:
The [file vendored](https://github.com/open-rmf/rmf_utils/blob/main/rmf_utils/include/rmf_utils/catch.hpp) by `rmf_utils` is indeed in the `rmf_utils/catch.hpp` path.
However, there is a `catch2` [package in rosdep](https://github.com/ros/rosdistro/blob/1f6d3a742ce750d3c0bb13797d8e1fc870a2d825/rosdep/base.yaml#L470-L485). We don't directly depend on it but if anyone in the dependency tree was it would be installed.
That package, in its [Ubuntu 22.04 version](https://packages.ubuntu.com/jammy/amd64/catch2/filelist) _provides_ a `catch2/catch.hpp` file, however in the [Ubuntu 24.04 version](https://packages.ubuntu.com/noble/amd64/catch2/filelist) it doesn't.
Hence in my opinion it was really working by pure chance and we should indeed include the vendored one

luca-della-vedova: WRT upgrading to `build_depend`, it seems there are two categories:

* Packages that use the `impl_ptr`, for which there is a `<depend>rmf_utils</depend>`
* Packages that only use rmf_utils for uncrustify, for which the find_package call is under a `if (BUILD_TESTING)` statement, those have it as a `<test_depend>`.

With this in mind do you still want to change them to a `build_depend`? They are not really needed unless we build tests

nexus_workcell_orchestrator/src/workcell_orchestrator.cpp
Yadunund: Sounds good.

nexus_common/src/action_client_bt_node_test.cpp
Yadunund: Thanks for explaining the reason for the header change.

According to [REP-0140](https://www.ros.org/reps/rep-0140.html#build-depend-multiple), I would say this should be a `build_depend` since `rmw_utils` is need at build time to supply the header needed for compilation.

Although under the `test_depend` description it's stated "When building with testing enabled, the <test_depend> packages are available for configuring and building the tests as well as running them." which suits are requirement.

I'm leaning towards `build_depend` here but let me know what you think based on `REP-0140`.

.github/workflows/nexus_workcell_editor.yaml
Yadunund: Got it.

Yadunund: Let's add a comment to explain that.

nexus_common/src/action_client_bt_node_test.cpp
luca-della-vedova: I went through the packages and made it a `build_depend` of the ones that include the `catch.hpp` header https://github.com/osrf/nexus/pull/35/commits/ca6ef25d2b36052dcc847bc5ee7e7a98865d613c

Yadunund (approved):

PR #37, opened 2024-10-30
Migrate to ros2dds bridge

### LINKED ISSUE #17: Switch to newer zenoh-plugin-ros2dds bridge
The new version of the bridge has several improvements but the one most interesting to us is the graph introspection https://discourse.ros.org/t/new-zenoh-bridge-for-ros-2/34163

luca-della-vedova:
Closes #17.

This seems to work but required a fair bit of changes in the `nexus_network_configuration` package, so we should probably rethink what that package does and how.

Specifically, it had a builtin support for generating configs from redf, but it actually didn't and just used hardcoded endpoints [here](https://github.com/osrf/nexus/blob/0741639171413cd077fb7862b85396210cc35bcb/nexus_integration_tests/config/zenoh/nexus_network_config.yaml#L10).
Furthermore, while with the `dds` plugin we can just specify endpoints as strings, the `ros2dds` plugin requires us to clearly say whether each substring is a publisher, subscriber, service server, service client, action server or action client.
The way I approached it in this PR is the same as the previous PR, [updating the config file](https://github.com/osrf/nexus/blob/ed94ad7f987ef516ee583423a3d80cb71ddbf68d/nexus_integration_tests/config/zenoh/nexus_network_config.yaml#L16-L27) accordingly, however there are a few important changes to consider

### Redf compatibility

`redf` does not have a way to specify for each topic / service etc. which nodes can publish and which can subscribe, so the only way to include `redf` config in its current implementation would be to allow each entity to be both, which breaks the isolation a bit since (for example), we currently do:

```
# System orchestrator
    allow:
      service_servers: [[...] "/register_workcell"]

# Workcell
    allow:
      service_clients: ["/register_workcell"]
```

Allowing both would mean _technically_ allowing a rogue workcell to advertise a `/register_workcell` client which could create a lot of trouble in the system.
Note however that this is also an issue in the current implementation.

Because there was no usage of `redf` for network configuration I couldn't test it and whether it worked or not before I'm fairly sure it will stop working with this PR.

### Queries timeout

The `ros2dds` bridge adds a concept of timeout for a query. They can be pretty [granular](https://github.com/eclipse-zenoh/zenoh-plugin-ros2dds/blob/329d7cf34331e430e7f8f2d0a870225ce608a8d9/DEFAULT_CONFIG.json5#L104-L125) but the main issue is that they also apply to actions so if an action was to take longer than the specified value it would fail.
I set it to 10 minutes as a default, but this should probably be double checked (increased? made more granular?).

### Namespaces

Sadly, while the ros2dds bridge supports namespaces, they currently [also apply them to global topics](https://github.com/eclipse-zenoh/zenoh-plugin-ros2dds/blob/329d7cf34331e430e7f8f2d0a870225ce608a8d9/DEFAULT_CONFIG.json5#L18-L26). This means that if we set a bridge with namespace `workcell_1` and asked it to allow a subscription to `/chatter` it would silently only allow `/workcell_1/chatter`. This makes it pretty painful because workcells need to access global services to register, so we can't rely on namespaces and have to add them manually to each workcell configuration.

### Warning spam

Last but not least, running the `ros2dds` bridge results in a lot of warnings being printed, they seem to be harmless but are still very noisy. An [upstream issue](https://github.com/eclipse-zenoh/zenoh-plugin-ros2dds/issues/290) has been opened and acknowledged by Zettascale.

### DIFF
--- a/nexus_integration_tests/config/zenoh/nexus_network_config.yaml
+++ b/nexus_integration_tests/config/zenoh/nexus_network_config.yaml
@@ -2,32 +2,68 @@
 # Use 'client' when there is an available Zenoh router, otherwise
 # use 'peer' for a distributed system.
 mode: peer
-# When true, discovery info is forwarded to the remote plugins/bridges
-forward_discovery: false
 # When true, activates a REST API used to administer Zenoh Bridge configurations
 enable_rest_api: true
-# Additional endpoints to allow, these could be endpoints not defined in REDF but are necessary for lifecycle transitions
-add_allowed_endpoints: ["/list_workcells", "/.*/workcell_state", "/.*/is_task_doable", "/.*/request", "/.*/queue_task", "/.*/remove_pending_task", "/register_workcell", "/register_transporter", ".*/available", ".*/transport", "/estop", "/.*/pause", "/.*/get_state", "/.*/change_state", "/.*/signal"]
 
 system_orchestrators:
-  - ros_namespace: system_orchestrator # ROS Namespace of the endpoints
+  - namespace: system_orchestrator # ROS Namespace of the endpoints
     # ROS Domain ID
-    domain_id: 14
+    domain_id: 0
     # Listening TCP Address of Zenoh bridge
     tcp_listen: ["0.0.0.0:7447"]
     # HTTP Port for the REST API
     rest_api_http_port: 8000
+    allow:
+      publishers: []
+      # TODO(luca) check if we need estop
+      subscribers: ["/estop", "/.*/workcell_state"]
+      service_servers: ["/list_workcells", "/register_workcell", "/register_transporter"]
+      # TODO(luca) check if transporter needs available endpoint
+      service_clients: ["/.*/queue_task", "/.*/remove_pending_task", "/.*/pause", "/.*/signal", "/.*/get_state", "/.*/change_state", "/.*/is_task_doable"]
+      action_servers: []
+      # TODO(luca) check if we really need transporter client
+      action_clients: ["/.*/request", "/.*/transport"]
+    queries_timeout:
+      default: 600.0
+
 
 workcell_orchestrators:
-  - ros_namespace: workcell_1
-    domain_id: 15
+  - namespace: workcell_1
+    domain_id: 1
     tcp_connect: ["0.0.0.0:7447"]
     rest_api_http_port: 8001
-  - ros_namespace: workcell_2
-    domain_id: 16
+    allow:
+      publishers: ["/workcell_1/workcell_state"]
+      subscribers: []
+      service_servers: ["/workcell_1/is_task_doable", "/workcell_1/queue_task", "/workcell_1/remove_pending_task", "/workcell_1/get_state", "/workcell_1/change_state", "/workcell_1/signal", "/workcell_1/pause"]
+      service_clients: ["/register_workcell"]
+      action_servers: ["/workcell_1/request"]
+      action_clients: []
+    queries_timeout:
+      default: 600.0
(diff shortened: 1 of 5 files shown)


nexus_integration_tests/config/zenoh/nexus_network_config.yaml
```diff
@@ -2,32 +2,68 @@
...
     rest_api_http_port: 8000
+    allow:
+      publishers: []
+      # TODO(luca) check if we need estop
+      subscribers: ["/estop", "/.*/workcell_state"]
+      service_servers: ["/list_workcells", "/register_workcell", "/register_transporter"]
+      # TODO(luca) check if transporter needs available endpoint
+      service_clients: ["/.*/queue_task", "/.*/remove_pending_task", "/.*/pause", "/.*/signal", "/.*/get_state", "/.*/change_state", "/.*/is_task_doable"]
+      action_servers: []
+      # TODO(luca) check if we really need transporter client
```
Yadunund, line 24: Any idea if these TODOs are still valid?

luca-della-vedova: The demo / integration test works well but the configuration of these endpoints depends on where nodes are ran / how complete we want this config file.
The main crux is that in this specific implementation the system orchestrator and the transporter node run in the same domain ID so _technically_ it is not necessary to allow all the transporter interfaces through Zenoh.
On the other hand, if we want to have this file be kind of a template that would also allow external transporters to connect to the system orchestrator we would need to add their interfaces here.

For the `estop` I'm just not 100% sure how it is supposed to work, since it seems it is a global topic do we just halt the whole system as soon as _one_ workcell decides to estop? That sounds a bit dangerous so I left a note here.

I'm erring on the side of "make this a template" at least until a proper redf integration for generation of config files is not done and permissively allow all the interfaces we might need but what do you think?

luca-della-vedova:
Warning spam was fixed upstream and in the bumped dependency https://github.com/osrf/nexus/pull/37/commits/20c2eb6f090d9d1221388bb0252b5a947efa8000

Yadunund (approved):
I'm unable to run a second job after the first one completes but that could be unrelated to the changes here. 

```
nexus_system_orchestrator-1] [ERROR] [1735196375.435664264] [system_orchestrator]: No workcell is able perform task [2]
[nexus_system_orchestrator-1] [ERROR] [1735196375.435700924] [system_orchestrator]: One or more tasks cannot be performed!
```

nexus_integration_tests/config/zenoh/nexus_network_config.yaml
Yadunund: > For the estop I'm just not 100% sure how it is supposed to work, since it seems it is a global topic do we just halt the whole system as soon as one workcell decides to estop? That sounds a bit dangerous so I left a note here.

Yeah we halt processes across all workcells if the estop is triggered. This was a requirement.

Yadunund:
I can confirm that with the `main` branch, I am able to run the `pick_from_conveyor` and `place_on_conveyor` processes back to back but am unable to do so with this branch.

luca-della-vedova:
Sigh. I bumped the patch version from 1.0.2 to 1.0.4 to avoid the warning spam and thought it was innocuous enough that I didn't have to retest the whole stack, but somehow that version (and the following 1.1.0) are broken.
I could reproduce, reverting it in https://github.com/osrf/nexus/pull/37/commits/d813ae98af56b35a7fa66ed544d546425a12945b helped, can you give it a spin?

Yadunund (approved):

PR #39, opened 2024-11-21
Migrate to UR robots

luca-della-vedova:
This PR migrates the tests to use UR robots, removing a significant chunk of code as well as the need to build all the `abb` packages from source, since the UR packages are distributed in the buildfarm.

It is pretty much ready for review, the only missing part is to remove the temporary building of the UR repo from source in the github action as soon as the next Jazzy sync happens, which should be in a few hours


luca-della-vedova:
Following feedback this PR should be changed to add a new workcell using UR robots alongside ABB robots. We can use `workcell_3` for this purpose.

Issue #41, opened 2024-12-11
More granular task capability checking for workcells


luca-della-vedova:
Nexus relies on the service [IsTaskDoable](https://github.com/osrf/nexus/blob/main/nexus_msgs/nexus_orchestrator_msgs/srv/IsTaskDoable.srv) to query which workcell can perform a specific task.
The task itself is made of both a plain string `type` and a YAML `payload`. However, [in the current implementation](https://github.com/osrf/nexus/blob/0741639171413cd077fb7862b85396210cc35bcb/nexus_workcell_orchestrator/src/workcell_orchestrator.cpp#L1019-L1030), the workcell orchestrator only checks whether a behavior tree exists with the requested `type` and doesn't do anything with the payload.

There might be cases where multiple workcells can perform the same type of task and the differentiating factor is in the payload itself, or cases in which a a workcell might be able to support a wide category of tasks but only a subset of payloads.
It should be possible to configure the function that checks whether a workcell can perform a task and implement more complex behaviors.

koonpeng:
Some context for this, there were some discussion on this. In general there are 2 paths that we can take

* A DIY approach where you have some table/database of which workcell can do which task with which payload etc. This should be low-code/no-code, users shouldn't have to compile anything.
* An automatic approach where each node/hardware driver can examine the payload and report if it can do the task. The biggest problem with this is that payloads must have a predefined schema, and there are too many variables for us to create a standard format.

These 2 approach are not mutually exclusive so we can have a mix of both.

Yadunund:
I'd like to rely on plugins which the orchestrator can load to decide whether the task can be performed. ie, plugins would implement a `can_perform_task()` like method. Checking whether the bt file exists in the local filesystem can be one such implementation. We can also define other plugins as described by @koonpeng above.

luca-della-vedova:
This should be fixed by #46

Issue #40, opened 2024-12-11
Customzable workcell cancellation behavior


luca-della-vedova:
Currently, trying to cancel a workcell task results in a NOOP if the task is being executed, and removing from the queue if it is not running yet. This behavior is [embedded](https://github.com/osrf/nexus/blob/0741639171413cd077fb7862b85396210cc35bcb/nexus_workcell_orchestrator/src/workcell_orchestrator.cpp#L291-L315) in the workcell orchestrator.

There might be types of workcells that can actually offer cancel a running task, so this should at least be configurable, two possible approaches:

* Add a cancel method to `Capability` that will let users define a cancellation behavior. https://github.com/osrf/nexus/blob/main/nexus_capabilities/include/nexus_capabilities/capability.hpp

* Allow users to specify a BT that the System / Workcell orchestrator should run upon cancellation.

koonpeng:
There was an unfinished feature that allows defining post-cancellation behavior on the BT. iirc the reason we ended with that solution is because

* If we stop the current actions asap, it will leave the workcell in a unknown / unrecoverable state. Then we would require a human to manually inspect things that are ok and reset the workcell.
* Cancelling can be very dangerous, imagine a pick and place involving a heavy payload, we need to know if it is safe to release the gripper etc. So I imagine that cancelling would most of the time be a no-op on the capability side.
* We can run the current actions to completion. Then we can assume that the workcell is still in a known state, so then we can have a BT to control how to reset the workcell. If the state is deemed unrecoverable then we should also cancel all queued tasks.

So the current behavior is the just easiest safe solution.

PR #42, opened 2024-12-11
Introduce a RMF transportation workcell

luca-della-vedova:
This PR introduces RMF integration into nexus, where RMF is a workcell, managed by the workcell orchestrator, that is capable of executing `transportation` tasks through a new [behavior tree](https://github.com/osrf/nexus/blob/5eb64499e1776b197fac68e533e9f4177ef35de6/nexus_integration_tests/config/rmf_bts/transportation.xml) and set of [capabilities](https://github.com/osrf/nexus/blob/5eb64499e1776b197fac68e533e9f4177ef35de6/nexus_capabilities/src/capabilities/transport_amr_capability.cpp).

This is in a simple demo stage. I brought in a [modified office world](https://github.com/osrf/nexus/blob/5eb64499e1776b197fac68e533e9f4177ef35de6/nexus_integration_tests/maps/office/office.building.yaml), with the only modifications being renaming the dispensers to the names of  the workcells, and added a new launch to `nexus_integration_tests` that launches RMF together with Nexus, as well as changed the movement of items to be based on an AMR rather than a mock transporter.

## Test it!

Clone, build and run:

```
ros2 launch nexus_integration_tests depot.launch.xml headless:=false
```

Submit a task:

```
ros2 action send_goal /system_orchestrator/execute_order nexus_orchestrator_msgs/action/ExecuteWorkOrder "{order: {id: '23', work_order: '$(cat config/pick_and_place.json)'}}"
```

You should see the transportation happening:

[Screencast from 2024-12-12 18-10-14.webm](https://github.com/user-attachments/assets/eb239c97-3904-4fbf-b84e-8a63d87cbd9e)

## PR breakdown

The PR is large but I'll try to condense the main decision (and potentially controversial) points I went through during the design.

### nexus_integration_tests vs nexus_demos

It would be more natural to create a new `nexus_demos` package that contains the bringup and I got halfway there before realising it would make the diff explode even further, so I went for an initial approach that reduces the number of changes in `nexus_integration_tests`, we can then do a followup PR that splits the package into a `nexus_demos` and a `nexus_integration_tests`, or maybe just rename it.

### Task cancellation

As noted in https://github.com/osrf/nexus/issues/40, the cancellation behavior of the workcell can't be customized and defaults to letting tasks run to completion. This means that the RMF task will not be cancelled and if a robot happens to be halfway through a long task and be waiting for a workcell that is cancelled, it will hang its waiting indefinitely. Once #40 is addressed we should add task cancellation to the `TransportAmr` capability.

### Is task doable / navgraph checking

As noted in https://github.com/osrf/nexus/issues/41, the payload can't be used for verifying task capability. Transportation tasks have a payload with a list of destinations and they will currently always return `true` regardless of whether the destinations exist or not. A more advanced capability checking that, for example, checks the fleet's navgraph for existence of waypoints, would be a better design.

### Map annotation

Visualizing the workcell requires [its position to be populated](https://github.com/osrf/nexus/blob/5eb64499e1776b197fac68e533e9f4177ef35de6/nexus_msgs/nexus_orchestrator_msgs/msg/WorkcellDescription.msg#L11-L15), however Nexus (and the workcell orchestrator) currently have no way to populate this information.
For now just for the sake of visualizing I wrote [a node that subscribes to the `/map` topic](https://github.com/osrf/nexus/blob/5eb64499e1776b197fac68e533e9f4177ef35de6/nexus_visualization/src/WorkcellStateVisualizer.cpp#L41-L80) and looks for all waypoints with the `pickup_dispenser` property and use their location to populate markers. It will then subscribe to states and update them.

A better long term design would involve passing the workcell orchestrator information about the location of the workcell, pass it to the `system orchestrator` when registering and refactoring the visualization node to regularly calls the `/list_workcells` service to query for existence of new workcells. I deferred this to avoid adding a large diff to the workcell orchestrator node and keep changes strictly addictive for review simplicity.

### Signaling

I introduced the capability of receiving signals for the `system orchestrator`, as well as change the default behavior tree to wait for the AMR _before_ starting the workcell, rather than halfway. This was done to improve reliability in case of parallel tasks (i.e. there is no risk of a workcell starting a task, just for the wrong AMR to come in) but parallel tasks are still not quite there so not sure if it is still needed. An example of behavior tree that implements this new logic is [here](https://github.com/osrf/nexus/blob/5eb64499e1776b197fac68e533e9f4177ef35de6/nexus_integration_tests/config/system_bts/pick_and_place.xml).

## What's next

Many things! But this PR is already at a very large size and I tried to keep the diff minimal (where I liberally define "diff" as pre-existing files that are changed and risk breaking existing behavior, not new additions that are more likely to be safe).

### Create a Gazebo simulation that includes workcells together with AMRs

Right now the workcells are not simulated in Gazebo, it would be great to have a proper simulation world so users can inspect what is happening.
Often these workcells have conveyor belts to feed the items to / from the AMRs, these would also be valuable additions.

### Simulate humans for workcells that are manually operated

In real life, not all workcells are automated and some are just operated by humans. We could mock this in simulation by just having a human in the dropoff point and a special behavior tree that just waits for an input.

### Task parallelism

Currently submitting parallel tasks _can_ risk deadlocking the system, since RMF and Nexus are somewhat independent. We should revisit the implementation to make sure we can have parallel tasks.

### SKU Tracking

It would be interesting to show the position and status of the SKUs in rviz. This is especially useful to know their state as they are being moved throughout the facility.

### Better handling of workcell location and registration

As noted in the `Map annotation` section of the PR description, populate the information at workcell registration time and not by subscribing to a `/map` topic.

### Post processing of waypoints for AMR tasks

Currently, whenever a work order is received, an AMR task that goes through all the workcells will be generated and each workcell will only be signaled to start when the AMR arrives.
This however, will be suboptimal in two corner cases:

1. If there is only one workcell and we don't want to use an AMR to transport, we will still request an AMR to the location which is unnecessary.
2. If there are multiple tasks being done by the same workcell, the AMR will have multiple "pickup" phases, although I believe this should be innocuous and just introduce some extra signaling.

It is actually a bit tricky to design a single behavior tree that works for all cases and I would actually suggest using a different behavior tree for different purposes, such as the first case.


nexus_capabilities/src/capabilities/transport_amr.hpp
```diff
@@ -0,0 +1,252 @@
+/*
```
Yadunund, line 1: Suggest to rename to `rmf_request`.

nexus_capabilities/src/capabilities/transport_amr.hpp
```diff
@@ -0,0 +1,252 @@
...
+#include <rmf_task_msgs/msg/api_request.hpp>
+#include <rmf_task_msgs/msg/api_response.hpp>
+
+#include <behaviortree_cpp_v3/action_node.h>
+
+#include <deque>
+
+namespace nexus::capabilities {
+
+struct AmrDestination {
```
Yadunund, line 38: Make `std::variant<rmf_dispenser_msgs::DispenserRequest, rmf_ingestor_msgs::IngestorRequest>` and rename to `Destination`?

Then we have separate capabilities to construct `DispenserRequest` and `IngestorRequest` from string and return a vector needed by the DispatchRmfRequest skill.

nexus_capabilities/src/capabilities/transport_amr.hpp
```diff
@@ -0,0 +1,252 @@
...
+#include <rmf_dispenser_msgs/msg/dispenser_result.hpp>
+#include <rmf_task_msgs/msg/api_request.hpp>
+#include <rmf_task_msgs/msg/api_response.hpp>
+
+#include <behaviortree_cpp_v3/action_node.h>
+
+#include <deque>
+
+namespace nexus::capabilities {
+
```
Yadunund, line 37: Add `rmf` namespace

nexus_capabilities/src/capabilities/transport_amr.hpp
```diff
@@ -0,0 +1,252 @@
...
+};
+
+/**
+ * Returns RUNNING until a signal is received from system orchestrator.
+ *
+ * Input Ports:
+ *   signal |std::string| Signal to wait for.
+ *   clear |bool| Set this to true to clear the signal when this node is finished.
+ */
+class DispatchRmfRequest : public BT::StatefulActionNode
```
Yadunund, line 51:
```suggestion
class DispatchRequest : public BT::StatefulActionNode
```

nexus_capabilities/src/capabilities/transport_amr_capability.hpp
```diff
@@ -0,0 +1,68 @@
...
+#include <rclcpp_lifecycle/lifecycle_node.hpp>
+
+#include <memory>
+
+namespace nexus::capabilities {
+
+/**
+ * Capability to plan robot arm motions.
+ */
+class TransportAmrCapability : public Capability
```
Yadunund, line 36:
```suggestion
class RMFRequestCapability : public Capability
```

nexus_integration_tests/config/office/tinyRobot_config.yaml
```diff
@@ -0,0 +1,48 @@
+# FLEET CONFIG =================================================================
```
Yadunund, line 1: Let's use `deliveryRobot` from RMF aka MiR.

nexus_system_orchestrator/src/dispatch_transporter.hpp
```diff
@@ -0,0 +1,91 @@
+/*
```
Yadunund, line 1: Double check if we need this.

Yadunund (changes requested):

nexus_capabilities/CMakeLists.txt
```diff
@@ -15,8 +15,11 @@ set(dep_pkgs
   nexus_common
   nexus_endpoints
   nexus_orchestrator_msgs
+  nlohmann_json
   rclcpp
   rclcpp_lifecycle
+  rmf_dispenser_msgs
+  rmf_task_msgs
   yaml-cpp
```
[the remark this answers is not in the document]
luca-della-vedova: https://github.com/osrf/nexus/pull/42/commits/6567fbb187c2da62c4c511ff26e5cbaaed2d8bc5

nexus_visualization/CMakeLists.txt
```diff
@@ -0,0 +1,77 @@
...
+  visualization_msgs
+)
+foreach(pkg ${dep_pkgs})
+  find_package(${pkg} REQUIRED)
+endforeach()
+
+#===============================================================================
+add_library(workcell_state_visualizer SHARED src/WorkcellStateVisualizer.cpp)
+
+ament_target_dependencies(workcell_state_visualizer
```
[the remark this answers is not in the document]
luca-della-vedova: https://github.com/osrf/nexus/pull/42/commits/c96fdbfb895c713e2a4282e81720613531f4d119

nexus_system_orchestrator/src/dispatch_transporter.hpp
luca-della-vedova: Actually yes! This is to execute this behavior tree node https://github.com/osrf/nexus/blob/c96fdbfb895c713e2a4282e81720613531f4d119/nexus_integration_tests/config/system_bts/main.xml#L8

What this system orchestrator node does is:

* Iterate over tasks in the work order
* Generate a transportation task for the current work order that visits all the workcells that have been assigned to each task.
* Bid this new transportation request to all the registered workcells. Fail if no workcell can do this transportation request.
* Output the transportation task as well as workcell that was assigned to it.

There is admittedly a bit of duplication in the bidding here and the bidding at the system orchestrator level. The problem that needs to be solved is that this specific task needs to have full knowledge of the result of the workcell task assignment to know where to send the AMRs, so it needs to run in a following step

nexus_integration_tests/config/office/tinyRobot_config.yaml
luca-della-vedova: @aaronchongth for the next iteration on sim world

nexus_capabilities/src/capabilities/transport_amr_capability.hpp
luca-della-vedova: https://github.com/osrf/nexus/pull/42/commits/ead8808d6abfe8d3bf95cc87dfb53c5c19d5b702

nexus_capabilities/src/capabilities/transport_amr.hpp
luca-della-vedova: https://github.com/osrf/nexus/pull/42/commits/ead8808d6abfe8d3bf95cc87dfb53c5c19d5b702

nexus_capabilities/src/capabilities/transport_amr.hpp
luca-della-vedova: https://github.com/osrf/nexus/pull/42/commits/ead8808d6abfe8d3bf95cc87dfb53c5c19d5b702

nexus_capabilities/src/capabilities/transport_amr.hpp
luca-della-vedova: https://github.com/osrf/nexus/pull/42/commits/ead8808d6abfe8d3bf95cc87dfb53c5c19d5b702

nexus_capabilities/src/capabilities/transport_amr.hpp
luca-della-vedova: Renamed https://github.com/osrf/nexus/pull/42/commits/18070e67ee1cfb1c4e1b9dfcf377cf1a2b4b647d.
I gave some thought to the `DispenserRequest` / `IngestorRequest` usage and I'm not 100% sure it is a good idea.

These structures are messages and using them would send the message (hah!) that they will be used in the stack, perhaps to be published.
The reality though, is that the main usage of this struct is to generate the task api request message, and the only use we have for dispenser requests is subscribing to them to tell the workcell that the robot arrived. We could do a separate structure and something like `std::variant<Pickup, Dropoff>`, where:

```
struct Pickup {
    std::string workcell;
}
// Same for dropoff
```

I would actually only implement `Pickup` in a first iteration since there is really no support for dropoffs in the current architecture so we can't even design a test to make sure they work. Hopefully though a `std::variant<Pickup>` with a well documented TODO to add more variants can be a good future extension point?

nexus_system_orchestrator/src/dispatch_transporter.hpp
luca-della-vedova: It could be renamed to something a bit clearer though, maybe `DispatchTransportationTask`? Or split into a `GenerateTransportTask` that only generates a task to visit all workcells and `DispatchTask` that takes it as an input and does the bidding / assignment

luca-della-vedova:
With https://github.com/osrf/nexus/pull/42/commits/2ddcafafd72c9565c97a599819796d4505dc1ddc I now fully duplicated the behavior trees between the old transporter based implementation and the new RMF based implementation. I also duplicated the integration test so that both are now part of our test suite

Yadunund:
> Is task doable / navgraph checking
As noted in https://github.com/osrf/nexus/issues/41, the payload can't be used for verifying task capability. Transportation tasks have a payload with a list of destinations and they will currently always return true regardless of whether the destinations exist or not. A more advanced capability checking that, for example, checks the fleet's navgraph for existence of waypoints, would be a better design.

With #41 merged, we could write a new `RMFTaskChecker` plugin that creates a bunch of `rmf_traffic::Graph()` objects based on input nav_graph paths and report whether a destination is connected to any graph. This can be a separate PR but let's open a ticket.

nexus_integration_tests/CMakeLists.txt
```diff
@@ -176,4 +176,79 @@ if(BUILD_TESTING)
...
+    DEPENDS ${output_nav_graphs_phony}
+  )
+
+  install(
+    DIRECTORY ${output_dir}
+    DESTINATION share/${PROJECT_NAME}/maps
+  )
+
+endforeach()
+
```
Yadunund, line 253: I'd really prefer committing the `.world` and `nav_graph` files needed to run the test as opposed to the usual RMF way of generating these files at build time. We can throw them inside `config/rmf/` along with anything else needed.
The world file just needs to be an empty floor with a couple robots included. Even better if the robot model URIs point to fuel.

We can follow the usual style when designing `nexus_demos`.

rmf.repos
```diff
@@ -0,0 +1,5 @@
+repositories:
+  rmf/rmf_demos:
+    type: git
+    url: https://github.com/open-rmf/rmf_demos.git
```
Yadunund, line 4: I don't think we should clone `rmf_demos` here. If we need the tinyRobot model, suggest committing it to Fuel with the slotcar plugin. See my comment above on committing the .world and nav_graph files.

Yadunund (changes requested):
Starting to look closer into the implementation. Sharing a big feedback to address while I review the capabilities. 
Let's not commit the various files needed to generate the office world and navgraphs and do so at compile time. Instead let's directly commit the .world and nav_graph files generated. Let's also throw all rmf related files inside `nexus_integration_tests/config/rmf/`.

nexus_capabilities/src/capabilities/transport_amr.hpp
Yadunund: > These structures are messages and using them would send the message (hah!) that they will be used in the stack, perhaps to be published.

Hmm I'm not sure if I agree with that. It's a data structure that we can reuse and avoid extra conversions between a class object and the message later on? Besides, all of this stuff is purely internal and we'll be relying on skills like `ExtractDestination` and `UnpackDestinationData` to create and parse the data structure.

nexus_visualization/src/WorkcellStateVisualizer.hpp
```diff
@@ -0,0 +1,84 @@
+/*
+ * Copyright (C) 2024 Open Source Robotics Foundation
```
Yadunund, line 2:
```suggestion
 * Copyright (C) 2025 Open Source Robotics Foundation
```

nexus_visualization/src/WorkcellStateVisualizer.hpp
```diff
@@ -0,0 +1,84 @@
+/*
```
Yadunund, line 1: Rename to `workcell_state_visualizer.*pp` to match rest of codebase.

nexus_visualization/src/WorkcellStateVisualizer.hpp
```diff
@@ -0,0 +1,84 @@
...
+#include <rmf_building_map_msgs/msg/building_map.hpp>
+#include <rmf_visualization_msgs/msg/rviz_param.hpp>
+
+#include <std_msgs/msg/color_rgba.hpp>
+#include <visualization_msgs/msg/marker_array.hpp>
+#include <visualization_msgs/msg/marker.hpp>
+
+#include <unordered_map>
+#include <unordered_set>
+
```
Yadunund, line 35: Add `nexus::visualization` namespace

nexus_visualization/src/WorkcellStateVisualizer.hpp
```diff
@@ -0,0 +1,84 @@
...
+    double x;
+    double y;
+    std::string map_name;
+
+    // Follows enum in WorkcellState.msg
+    // 0 indefinite, 1 idle, 2 busy
+    uint8_t status = 0;
+    std::string task_id = "<None>";
+    std::string message = "<None>";
+  };
```
Yadunund, line 62: Let's not create a struct with new fields but reuse message definitions in nexus. Parse BuildingMap and create the `description` ptr and initialize `state` to `nullptr`. Then in `state_subscription_cb`, simply update the `state` ptr.

```suggestion
  struct Workcell {
     std::shared_ptr<const nexus_orchestration_msgs::msg::WorkcellDescription> description;
     std::shared_ptr<const nexus_orchestration_msgs::msg::WorkcelState> state;
  };
```

nexus_visualization/src/WorkcellStateVisualizer.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+  void initialize_state_subscriptions();
+
+  void timer_cb();
+
+  std::string _current_level;
+
+  rclcpp::TimerBase::SharedPtr _publish_timer;
```
Yadunund, line 70: No timers please. Let's make the publication event based. The QoS.durability can be `transient_local`. Publish a marker only if the state changed.

nexus_visualization/src/WorkcellStateVisualizer.hpp
```diff
@@ -0,0 +1,84 @@
...
+  rclcpp::Subscription<BuildingMap>::SharedPtr _map_sub;
+  rclcpp::Subscription<RvizParam>::SharedPtr _param_sub;
+  std::vector<rclcpp::Subscription<nexus::endpoints::WorkcellStateTopic::MessageType>::SharedPtr> _state_subs;
+  rclcpp::Publisher<MarkerArray>::SharedPtr _marker_pub;
+
+  // Map workcell name to a unique marker id
+  std::unordered_map<std::string, std::size_t> _ids;
+  std::size_t _next_available_id;
+
+  std::unordered_map<std::string, WorkcellState> _workcell_states;
```
Yadunund, line 80:
```suggestion
  std::unordered_map<std::string, Workcell> _workcell_states;
```

Yadunund (changes requested):
Feedback on workcell visualizer.

luca-della-vedova:
> Starting to look closer into the implementation. Sharing a big feedback to address while I review the capabilities. Let's not commit the various files needed to generate the office world and navgraphs and do so at compile time. Instead let's directly commit the .world and nav_graph files generated. Let's also throw all rmf related files inside `nexus_integration_tests/config/rmf/`.

Note that we will _still_ need to commit both the building and the navgraph if we want to run simulation so will likely have to still commit both original map and generated files.
The reason is that the simulated robots use the map published by the `building_map_server` to find what level they are in.

rmf.repos
luca-della-vedova: Apart from the launch files that I guess we can duplicate, we need `rmf_demos_fleet_adapter` to manage the slotcar robots. Unless we copy it in (!) or create a much more bare simulation method we can't really get away from having `rmf_demos` cloned.
If we distributed at least `rmf_demos_fleet_adapter` on the buildfarm we could remove this dependency.

luca-della-vedova: Or we could also do the nuclear option of not using the `rmf_demos_fleet_adapter` but using the very old C++ `full_control` fleet adapter but that's also not a great idea imo

nexus_visualization/src/WorkcellStateVisualizer.hpp
luca-della-vedova: https://github.com/osrf/nexus/pull/42/commits/851082c04d074d28fd869524c448cbff0468c94c

nexus_visualization/src/WorkcellStateVisualizer.hpp
luca-della-vedova: https://github.com/osrf/nexus/pull/42/commits/851082c04d074d28fd869524c448cbff0468c94c

nexus_visualization/src/WorkcellStateVisualizer.hpp
luca-della-vedova: https://github.com/osrf/nexus/pull/42/commits/64abc78ad8a1bafb51d64d76f09689bd9a4e736b

nexus_visualization/src/WorkcellStateVisualizer.hpp
luca-della-vedova: https://github.com/osrf/nexus/pull/42/commits/64abc78ad8a1bafb51d64d76f09689bd9a4e736b

nexus_visualization/src/WorkcellStateVisualizer.hpp
luca-della-vedova: https://github.com/osrf/nexus/pull/42/commits/64abc78ad8a1bafb51d64d76f09689bd9a4e736b

rmf.repos
luca-della-vedova: What I have done for now is still clone it (since we need the fleet adapter) but remove all other `rmf_demos` dependencies.
Since the CI should do a build "up to" `nexus_integration_tests` the total impact should be only building one Python package which is negligible.
I am not sure a better option exists, other than writing a full mock fleet adapter with a mock simulation, using legacy fleet adapters or distributing binaries of the demo fleet adapter, as mentioned above.

luca-della-vedova:
I addressed the feedback on the visualization node and spun it out into another PR https://github.com/osrf/nexus/pull/57, marked as draft since it should only be merged after this (and probably after demos as well)

rmf.repos
Yadunund: Fair enough. We should look into releasing `rmf_demos` binaries for Jazzy...

luca-della-vedova:
In https://github.com/osrf/nexus/pull/42/commits/f55bab26bbdebd4678aed969ed21efbd36276e7a I reverted the signaling at the system orchestrator level.
Now the workcell behavior trees are exactly the same regardless of whether it is a pick and place on a conveyor or on an AMR.
Sadly remapping gets in the way since we need to make sure we run the _same_ workcell behavior tree but a _different_ system orchestrator behavior tree (and not only the `main.xml`, but also the one that is loaded, `pick_and_place.xml`).

Furthermore, I explored the idea of removing all the duplicated behavior trees / work orders altogether in https://github.com/osrf/nexus/pull/42/commits/f9705a7aa7d30170382f58d851164c271e08cd83. The idea is that if we just expose the `remap_task_types` parameter to the launch files we can use it to make sure the same work order results in a different system orchestrator behavior tree but the same workcell behavior tree. But happy to revert it if this is not desirable

aaronchongth:
Although integration tests were passing, I noticed a weird behavior when running non-headless. `workcell_2` seems to start moving just as the AMR starts leaving `workcell_1`. [CI shows the same thing too](https://github.com/osrf/nexus/actions/runs/12839744332/job/35807464813#step:10:7238)

I've managed to narrow it down to [this flag](https://github.com/osrf/nexus/pull/42/files#diff-bf34574ad65f13c9b6b7360fa8a4fd76a269cb249a92a15df8adaef29d81dfe1R224),  where despite the initialization to `false`, it appears to be `true` when `WaitForAmr::onStart` gets triggered.

This can be verified by checking out https://github.com/osrf/nexus/pull/42/commits/5d3daf82104d10f7bf46791f11fb73f59c2c601a, and commenting out [this line](https://github.com/osrf/nexus/pull/42/commits/5d3daf82104d10f7bf46791f11fb73f59c2c601a#diff-735eaa7b4dba1610ed38c668db00a853a89e152bd05b7d9e93ce7fc9ef0c0b2dR274). The printouts will show that the `_amr_ready` is already `true` despite no dispenser requests sent. Watch out for

```
[nexus_workcell_orchestrator-14] [INFO] [1737357565.631919038] [rmf_nexus_transporter]: CHECKING: workcell [workcell_2], rmf_task_id [compose.dispatch-9ccd185db2], amr_ready: [amr ready]
```

Other than this line, everything else seem good

aaronchongth:
Per discussion, there may exist an issue regarding multiple work orders at the same time. After investigating more and trying it out, I found that they were working as expected :thinking:, with some observations,

```bash
cd nexus_integration_tests

# First order accepted
ros2 action send_goal /system_orchestrator/execute_order nexus_orchestrator_msgs/action/ExecuteWorkOrder "{order: {id: '23', work_order: '$(cat config/pick_and_place.json)'}}"

# While the first order is still being executed, run a second same order with a different order ID
ros2 action send_goal /system_orchestrator/execute_order nexus_orchestrator_msgs/action/ExecuteWorkOrder "{order: {id: '24', work_order: '$(cat config/pick_and_place.json)'}}"
# This gets rejected/aborted, saying failed to assign task to workcells as the task ID [2] already exists (this refers to the step ID)

# Save this new order somewhere, https://gist.github.com/aaronchongth/b5b92f140d539c33e0d0ec23b414d70c
# This new order just modifies the step IDs to 3.0 and 4.0
# While the first order is still being executed, send this new order
ros2 action send_goal /system_orchestrator/execute_order nexus_orchestrator_msgs/action/ExecuteWorkOrder "{order: {id: '24', work_order: '$(cat config/new_pick_and_place.json)'}}"
# This order gets accepted and starts being executed after order ID 23 is done

# When the order ID 23 has been completed, send in the same original order, with a different order ID,
ros2 action send_goal /system_orchestrator/execute_order nexus_orchestrator_msgs/action/ExecuteWorkOrder "{order: {id: '25', work_order: '$(cat config/pick_and_place.json)'}}"
# Order gets accepted, and only starts after order ID 24 is done
```

I haven't been able to replicate the behavior we discussed about, regarding work orders interfering with each other during completion. However summarizing some observations,
* when a work order is being executed, sending in more work orders with the same step ID will be rejected (the above scenario)
* when a work order is being executed, sending in work orders with different step IDs will be accepted
* after the a work order has been completed, sending in the same work order with a different order ID containing same step IDs will be accepted
* [step ID is parsed as double, but retrieved as int](https://github.com/osrf/nexus/blob/main/nexus_common/include/nexus_common/models/work_order.hpp#L71-L74), this causes step IDs of 1.0 and 1.1 to be treated the same and gets rejected. Is there a reason we use doubles but parse as int?

Yadunund:
> Per discussion, there may exist an issue regarding multiple work orders at the same time. After investigating more and trying it out, I found that they were working as expected 🤔, with some observations,
> 
> ```shell
> cd nexus_integration_tests
> 
> # First order accepted
> ros2 action send_goal /system_orchestrator/execute_order nexus_orchestrator_msgs/action/ExecuteWorkOrder "{order: {id: '23', work_order: '$(cat config/pick_and_place.json)'}}"
> 
> # While the first order is still being executed, run a second same order with a different order ID
> ros2 action send_goal /system_orchestrator/execute_order nexus_orchestrator_msgs/action/ExecuteWorkOrder "{order: {id: '24', work_order: '$(cat config/pick_and_place.json)'}}"
> # This gets rejected/aborted, saying failed to assign task to workcells as the task ID [2] already exists (this refers to the step ID)
> 
> # Save this new order somewhere, https://gist.github.com/aaronchongth/b5b92f140d539c33e0d0ec23b414d70c
> # This new order just modifies the step IDs to 3.0 and 4.0
> # While the first order is still being executed, send this new order
> ros2 action send_goal /system_orchestrator/execute_order nexus_orchestrator_msgs/action/ExecuteWorkOrder "{order: {id: '24', work_order: '$(cat config/new_pick_and_place.json)'}}"
> # This order gets accepted and starts being executed after order ID 23 is done
> 
> # When the order ID 23 has been completed, send in the same original order, with a different order ID,
> ros2 action send_goal /system_orchestrator/execute_order nexus_orchestrator_msgs/action/ExecuteWorkOrder "{order: {id: '25', work_order: '$(cat config/pick_and_place.json)'}}"
> # Order gets accepted, and only starts after order ID 24 is done
> ```
> 
> I haven't been able to replicate the behavior we discussed about, regarding work orders interfering with each other during completion. However summarizing some observations,
> 
> * when a work order is being executed, sending in more work orders with the same step ID will be rejected (the above scenario)
> * when a work order is being executed, sending in work orders with different step IDs will be accepted
> * after the a work order has been completed, sending in the same work order with a different order ID containing same step IDs will be accepted
> * [step ID is parsed as double, but retrieved as int](https://github.com/osrf/nexus/blob/main/nexus_common/include/nexus_common/models/work_order.hpp#L71-L74), this causes step IDs of 1.0 and 1.1 to be treated the same and gets rejected. Is there a reason we use doubles but parse as int?

Thanks for investigating further.

- I do think the current behavior where Step IDs also need to be unique is not ideal. Let's open a ticket and update behavior? The Work Order ID should be unique. The `id` we pass to the workcell in the [WorkcellTask](https://github.com/osrf/nexus/blob/7c6ebf7851b90d0683583e745012809f8f5ae0d5/nexus_msgs/nexus_orchestrator_msgs/msg/WorkcellTask.msg#L1) request can be a unique combination of the Work Order ID and the Step ID.
- We should parse the step ID as an integer and not float (and update work order defns)
- Regarding the parallel behavior:
  - Even with unique Step IDs, it seems one Work Order needs to complete execution before another can begin. This is not the case with the conveyor plugin. If a workcell needed for the second order is available, it should be able to process the step for the second order. I just tried submitting two separate jobs (unique Step IDs) with `use_rmf_transporter:=False` and the system executes both in parallel, ie While Step 2 in Work Order 1 is being processed by workcell_2, workcell_1 is processing Step 1 in Work Order 2. I believe this is not possible with the RMF integration given the BT definition and might be related to the point about signaling that Luca made above. 
  - I would expect a second RMF bid to go out which gets assigned to the second AMR but the AMR is dispatched only when the required workcell is available. 
  - Could you check if there is a quick fix by modifying the BTs? Else let's open a ticket and tackle this in a follow up PR.

aaronchongth:
Gotcha, I was actually looking into the issue that you showed me, when there were 2 work orders in parallel, completion of the first work order, somehow completed the second work order as well. Unless I remembered wrongly.

> I would expect a second RMF bid to go out which gets assigned to the second AMR but the AMR is dispatched only when the required workcell is available.

 Thanks for flagging out this other parallel work order scenario. Yeah that is happening due to how the [RMF transportation workcell's BT](https://github.com/osrf/nexus/pull/42/files#diff-ccaf5bcbed419f22c5997f33eb0d7f24c3cf747149df67ccc4e97ef567f16833) is currently designed, where the root BT takes care of dispatching an RMF task as well as keeping track of each looped destination in the RMF task (to handle the dispenser requests). IIUC, this means the RMF transportation workcell is never "done" until the whole work order is completed, before dispatching another robot.

From the `Signalling` section in this PR's description, it looks like this is by design to prevent the wrong AMR from reaching the workcell that is waiting for another AMR. Unfortunately I can't think of a way to resolve this on a BT level confidently. I have some ideas, will open a ticket for this particular situation to discuss more.

nexus_system_orchestrator/src/dispatch_transporter.hpp
Yadunund: Perhaps `MakeRMFTransportRequest` is more appropriate.

Als I've opened https://github.com/osrf/nexus/issues/66 to address bidding duplication.

I guess if the constructor for `DispatchTransporter` accepted a raw_ptr/reference to the `SystemOrchestrator` we could invoke (_assign_workcell_task](https://github.com/osrf/nexus/blob/7c6ebf7851b90d0683583e745012809f8f5ae0d5/nexus_system_orchestrator/src/system_orchestrator.cpp#L1000) here.

I'll add a TODO in the code and we can revisit this.

Yadunund (approved):
Thanks for trailblazing the approach of integrating RMF as a workcell that provides transportation services!

The experience here has been invaluable in better understanding the pros and cons of this approach vs integration via `nexus_transporter`. The main issue uncovered is that we can't run multiple work orders in parallel since workcells don't have the ability to run tasks in parallel yet. Further, the implementation here implicitly defines the `task.type` for workcells that perform transportation services, ie, `task.type = transportation` with an internal schema for the task params to include destinations/pickups etc. Lastly, all transportation steps required for a job is performed by the same transporter workcells. However in practice, we might distribute this among different transporters (eg. conveyor for some segments, AMRs for others (or even different AMRs)).
I've opened a meta-ticket to track things we need to implement to better support Workcells as transporters in general. https://github.com/osrf/nexus/issues/67

For now we can merge this PR in and iterate in subsequent PRs.

Issue #44, opened 2024-12-30
Adding Apache-2.0 license


vmbrasseur:
Hey there, team.

It seems there isn't a license in this repo yet. I suspect that's just an oversight as everyone's busy with the other work.

Does anyone mind whether I add Apache-2.0 in here for you?

Yadunund:
Hi @vmbrasseur,

That would be much appreciated!

vmbrasseur:
Happy to help! There's a PR hanging out for whenever someone has time.

PR #45, opened 2024-12-31
Adding Apache-2.0 license

vmbrasseur:
Per #44, this license file hadn't been added yet.

This PR fixes that by adding the Apache-2.0 license, which is in line with other OSRF projects. :-)


Yadunund (approved):
LGTM! Thanks for adding this!

PR #47, opened 2024-12-31
Make main behavior tree configurable

luca-della-vedova:
This PR makes the `main` behavior tree configurable through a parameter since it was found in #42 that it is not generic enough that it can be used to execute different classes of work orders.

Overview:

* A new parameter `main_bt_filename` has been introduced, which takes just the name of a behavior tree that is contained in the `bt_path` parameter.
* The `bt_path` system orchestrator parameter now just points to the folder containing behavior trees.
* The main behavior tree to be used will just be a path concatenation of `bt_path` and `main_bt_filename`.
* The `main_bt_filename` parameter is editable and a parameter callback has been added. Trying to set it to a non existing file will fail and return an error to the user.


nexus_integration_tests/launch/control_center.launch.py
```diff
@@ -132,6 +132,7 @@ def launch_setup(context, *args, **kwargs):
                     """{
                         pick_and_place: [place_on_conveyor, pick_from_conveyor],
                     }""",
+                "main_bt_filename": "main.xml",
```
Yadunund, line 135: Please `DeclareLaunchArgument` and get the value via `LaunchConfiguration` instead of hardcoding here. That way, parent launch files like `launch.py` can overwrite the value if needed.

Yadunund (changes requested):
LGTM except for one minor thing.

nexus_integration_tests/launch/control_center.launch.py
luca-della-vedova: Ah totally forgot, https://github.com/osrf/nexus/pull/47/commits/99c0c6902004dbbc1bae4cd347cd43b93de41069. It's not declared in the root `launch.py` but it is still passed to the `control_center.launch.py`, I can declare it in both places if it makes it clearer

Yadunund (approved):

PR #49, opened 2025-01-06
`nexus_demos` package

aaronchongth:
* splitting off what @luca-della-vedova set up within `nexus_integration_tests` previously, into a new package, including all the bts and config files under the `config` directory
* use depot world now, https://app.gazebosim.org/OpenRobotics/fuel/models/Depot 
* depot scan is from https://github.com/ros-navigation/navigation2/blob/main/nav2_bringup/maps/depot.pgm
* removed office world
* added custom cyclonedds config and some instructions to use it if the participant ID can't be created
* commits the generated `.world` file and navigation graphs directly into the package, with instructions in README on how to regenerate them. Building the `nexus_demos` package will however still need the model downloading step, since the models are still added into `.building.yaml` from `traffic-editor` (unless we want to eyeball and add assets into the template world file instead)

Screenshot_20250109_150356

## Testing

```
# build
colcon build --packages-up-to nexus_demos

# launch
ros2 launch nexus_demos depot.launch.xml headless:=0

# submit work order
cd ~/ws_nexus/src/nexus_demos
ros2 action send_goal /system_orchestrator/execute_order nexus_orchestrator_msgs/action/ExecuteWorkOrder "{order: {id: '23', work_order: '$(cat config/pick_and_place_rmf.json)'}}"
```

## Things to consider (with subsequent PRs)

* perhaps migrating most of the launch files, rviz files,  out of `nexus_integration_tests` and into `nexus_demos` to be made more reusable, so that the integration tests are dependent on the demos, not the other way around


luca-della-vedova:
> Requires [open-rmf/rmf_traffic_editor#523](https://github.com/open-rmf/rmf_traffic_editor/pull/523), to ensure simulation world camera initializes inside the factory. For testing now, before it gets merged, remember to checkout `rmf_traffic_editor` to `ac/skip-camera-pose-flag`

Currently this branch uses `jazzy` binaries, this means that we will need to change to either:
* Build RMF from source from now on.
* Build from source, backport the "minor" feature to `jazzy` and re-release then move back to binaries.

I'm a bit hesitant to build from source since it would add a good chunk of time to our CI.
We could _probably_ get a way with only building one package from source but that is a very dangerous route to go in, so we would most likely need to build the whole RMF stack.
The `rmf_demos` CI can be a good benchmark of the additional build time since it builds up to rmf_demos, which seems to be ~30 mins, so that would likely bring the CI from 20 min to 50 min (!).
A last possible alternative is to keep relying on binaries for CI but source builds for local, but this is also not very great.
What do you think? I wonder if it can be "sidestepped" by some (sigh) careful traffic editor map design that would make the builtin camera pose estimation do something sensible? Maybe adding a floor element to the level?

aaronchongth:
I was planning to set up docker images for underlay workspaces (RMF and zenoh bridge), and rely on nightly builds to set up the images. Then in our CI we just use those images to continue nexus builds and tests, that should speed things up

I prefer to stick to the current approach of using binaries as much as possible, and only use source builds if necessary. I'll open another PR targeting https://github.com/osrf/nexus/pull/42, when I'm done. Let me know what you think.

luca-della-vedova:
Upstream PR https://github.com/open-rmf/rmf_traffic_editor/pull/523 merged. I'm still a bit on the fence on whether to hack it on traffic editor or introduce an RMF partial source build, I'll defer it to the result of the docker experimentation

aaronchongth:
Per discussion, this PR has diverged, and most changes will wait for https://github.com/osrf/nexus/pull/42
The demo map, discussion points and feedback will be applied on https://github.com/osrf/nexus/pull/61

PR #50, opened 2025-01-06
Pin working version of rclrs

Yadunund:


luca-della-vedova:
I think the fix will be more involved and will require either upstream changes to the `rmf_site` repo, or committing a lockfile to this repo.

luca-della-vedova (approved):
https://github.com/osrf/nexus/pull/50/commits/0ac6bfa42945420f2b5948ec937439c8f340c6de "fixed" it, for the time being.

Yadunund:
Ah thanks for the actual fix!

Issue #48, opened 2025-01-03
Unregister a workcell


Yadunund:
- Add srv definition to `UnregisterWorkcell` : https://github.com/osrf/nexus/tree/main/nexus_msgs/nexus_orchestrator_msgs/srv
- Implement service server in system orchestrator and update task mechanisms accordingly
- Implement a function in the workcell orchestrator to make a request to unregister itself. Call this method in the destructor and/or within `on_deactivate`

PR #52, opened 2025-01-07
Fix registration logic in lifecycle manager

luca-della-vedova:
This PR works toward https://github.com/osrf/nexus/issues/48 by fixing corner cases with workcells bringup / bringdown.
The logic in the lifecycle manager was flawed, specifically in the [branch](https://github.com/osrf/nexus/blob/a0eff3c52dcf63ad499da0f88daa2df9bceed220/nexus_lifecycle_manager/include/nexus_lifecycle_manager/lifecycle_manager.hpp#L373-L403) where nodes already exist and a new one is added.

Two main issues are fixed by this PR.

### Re-registering the same workcell.

If Nexus has a single workcell, that workcell is brought down without proper cleanup (i.e. it suddenly crashes) then it's readded, we will go into the branch I linked above. However, when we try to get the target state https://github.com/osrf/nexus/blob/a0eff3c52dcf63ad499da0f88daa2df9bceed220/nexus_lifecycle_manager/include/nexus_lifecycle_manager/lifecycle_manager.hpp#L376-L381

We will actually query the workcell we are trying to create right now, hence try to transition it to its current state ending up in a NOOP. The new workcell will be successfully added but remain unconfigured.

### `workcell_1` is created, `workcell_1` is killed, `workcell_2` is created

Another corner case of the check above, now we will actually try to get the target state from a workcell that died, was not cleaned up properly and will thus be unreachable, adding new nodes will fail.

## The solution

The solution I came up with is actually a large simplification of the logic. I just created a new class variable `_target_state` that contains the state that we want the nodes to be in.
* If the lifecycle manager is set to `autostart` the target state will be `ACTIVE`, otherwise `UNCONFIGURED`.
* `_autostart` itself can be removed, now it just sets `_target_state` to ACTIVE instead.
* `system_active` can be removed, we just check that the `_target_state` is ACTIVE.
* Whenever we do a `changeStateForAllNodes` we will also update this internal `_target_state`.
* Whenever a new node is added, we transition it to `_target_state`.

## The drawback

The only "corner" case I can think of is in cases in which we want to manage a set of nodes at different states, in which the concept of a single target state that new nodes should be transitioned to will not be applicable. Two main points about this:

* Most importantly I would argue that the previous version was also broken. We would arbitrarily take the state the first item in an `unordered_map` and transition new nodes to that state. `unordered_map` is... unordered... so the first item could be literally any node depending on what the hashing does.
* I also feel the lifecycle manager's main scope is to make sure that all the managed nodes are at the same target state. If that is the case what I would actually recommend is only exposing APIs that transition _all_ the nodes to a state and make APIs that transition a single node private, so it won't be possible for external users to create invalid internal state.

### DIFF
--- a/nexus_lifecycle_manager/include/nexus_lifecycle_manager/lifecycle_manager.hpp
+++ b/nexus_lifecycle_manager/include/nexus_lifecycle_manager/lifecycle_manager.hpp
@@ -118,6 +118,9 @@ public:
       if (!state.has_value())
       {
         ok = false;
+        RCLCPP_WARN(
+          this->node_->get_logger(), "Failed to get state for node [%s]",
+          name.c_str());
         continue;
       }
 
--- a/nexus_lifecycle_manager/include/nexus_lifecycle_manager/lifecycle_manager.hpp
+++ b/nexus_lifecycle_manager/include/nexus_lifecycle_manager/lifecycle_manager.hpp
@@ -89,8 +89,8 @@ public:
 
   bool _system_active{false};
   std::chrono::seconds _service_request_timeout{10s};
-  // autostart, if set to true, will transition the first added node to ACTIVE
-  bool _autostart{false};
+  // Lifecycle state that managed nodes should be in
+  uint8_t _target_state = 0;
 
   std::shared_ptr<std::thread> spin_thread_;
   // Callback group used by services
@@ -123,6 +123,7 @@ public:
         transitionGraph.atGoalState(state.value(), transition) ||
         client->change_state(transition));
     }
+    this->_target_state = transition;
     return ok;
   }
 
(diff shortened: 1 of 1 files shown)


nexus_lifecycle_manager/include/nexus_lifecycle_manager/lifecycle_manager.hpp
```diff
@@ -117,11 +116,14 @@ class Implementation
     {
       std::optional<uint8_t> state = client->get_state();
       if (!state.has_value())
+      {
         ok = false;
+        continue;
```
Yadunund, line 121: can we log a warning telling users that we were unable to get the state for this client?
Second should we add this client to a list and then remove these clients outside this loop?

Yadunund (changes requested):
Thanks for working on this! The approach is a lot better. Left a minor comment to consider before we can merge this.

nexus_lifecycle_manager/include/nexus_lifecycle_manager/lifecycle_manager.hpp
```diff
@@ -117,11 +116,14 @@ class Implementation
...
+      }
 
-      ok = ok && (
-        transitionGraph.atGoalState(state.value(), transition) ||
-        client->change_state(transition));
+      ok &= transitionGraph.atGoalState(state.value(), transition) ||
```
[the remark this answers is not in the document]
luca-della-vedova: This is an interesting one, the idea behind this change is that previously as soon as `ok` became false the short circuiting would make all the following clients skip transitioning as well.
With the new logic without short circuiting instead, all the following client will still transition.
This is also why I added the `continue` above, before the short circuiting would make sure we never call `.value()` even as new nodes do indeed have a state but since short circuiting is removed we need an early guard clause

nexus_lifecycle_manager/include/nexus_lifecycle_manager/lifecycle_manager.hpp
luca-della-vedova: https://github.com/osrf/nexus/pull/52/commits/967fade404187ce19724bd929b71588bcdf89cb6.
WRT the removing not sure, I noticed that there is an API to remove nodes, it's just that it's not used anywhere in the codebase https://github.com/osrf/nexus/blob/c3200d5f2abdd57339699e461b41b552a9d0e163/nexus_lifecycle_manager/include/nexus_lifecycle_manager/lifecycle_manager.hpp#L512-L516

I'd like to try to introduce it first to do proper cleanup (i.e. with unregistration) rather than "silently" cleaning up automatically. Nexus tends to have _a lot_ of output in the terminal and I fear a single one off warning during bringup might be missed otherwise, and lead to more difficult to debug issued down the line

nexus_lifecycle_manager/include/nexus_lifecycle_manager/lifecycle_manager.hpp
```diff
@@ -117,11 +116,17 @@ class Implementation
     {
       std::optional<uint8_t> state = client->get_state();
       if (!state.has_value())
+      {
         ok = false;
+        RCLCPP_WARN(
+          this->node_->get_logger(), "Failed to get state for node [%s]",
```
Yadunund, line 122: Nit
```suggestion
          this->node_->get_logger(), "Failed to get state for node [%s]. Skipping state change...",
```

Yadunund (changes requested):

nexus_lifecycle_manager/include/nexus_lifecycle_manager/lifecycle_manager.hpp
Yadunund: Fair enough.

Yadunund (approved):

PR #55, opened 2025-01-08
Introduce a common TaskRemapper with wildcard support

luca-della-vedova:
This PR refactors the task remapping capability in `nexus_system_orchestrator` (which was fully fledged) and the one in `nexus_workcell_orchestrator` (that was actually unimplemented) into a common `TaskRemapper` class.

This class has basic wildcard support, if it finds such a remapping:

```
    pick_and_place: ["*"],
```

It will override every other defined remap and just remap every single task to the requested `pick_and_place`.
I added a set of unit tests to show how the functionality works and now both system and workcell orchestrator share the same codebase for task remapping.

In the future we can consider more complex behaviors (i.e. proper wildcard support)

### DIFF
--- a/nexus_common/include/nexus_common/task_remapper.hpp
+++ b/nexus_common/include/nexus_common/task_remapper.hpp
@@ -22,6 +22,9 @@
 
 #include <rclcpp/rclcpp.hpp>
 
+#include <optional>
+#include <string>
+
 namespace nexus::common {
 
 /**
@@ -37,8 +40,9 @@ public:
 
   /*
    * Remaps, if necessary, the input task
+   * Returns a value if the task was remapped, std::nullopt otherwise
    */
-  std::string remap(const std::string& task) const;
+  std::optional<std::string> remap(const std::string& task) const;
 
 private:
   // If present, match every incoming task to the target task
--- a/nexus_common/include/nexus_common/task_remapper.hpp
+++ b/nexus_common/include/nexus_common/task_remapper.hpp
@@ -22,6 +22,8 @@
 
 #include <rclcpp/rclcpp.hpp>
 
+#include <yaml-cpp/yaml.h>
+
 #include <optional>
 #include <string>
 
@@ -34,9 +36,18 @@ class NEXUS_COMMON_EXPORT TaskRemapper
 {
 public:
   /*
-   * Initialize the remapper with the value of a ROS parameter containing a YAML
+   * Initialize the remapper with the value of a ROS parameter containing a YAML.
+   *
+   * The YAML should have a sequence of key:value types where the key is the task to
+   * remap to, and the value is a list of tasks to remap to the key.
+   * For example:
+   *
+   *    pick_and_place: [pick, place]
+   *    a_and_b: [a, b]
+   *
+   * Will remap "pick" or "place" to "pick_and_place", "a" and "b" to "a_and_b"
    */
-  TaskRemapper(const std::string& param);
+  TaskRemapper(const YAML::Node& remaps);
 
   /*
    * Remaps, if necessary, the input task
--- a/nexus_common/include/nexus_common/task_remapper.hpp
+++ b/nexus_common/include/nexus_common/task_remapper.hpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2024 Open Source Robotics Foundation
+ * Copyright (C) 2025 Open Source Robotics Foundation
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/nexus_common/include/nexus_common/task_remapper.hpp
+++ b/nexus_common/include/nexus_common/task_remapper.hpp
@@ -1,0 +1,51 @@
+/*
+ * Copyright (C) 2024 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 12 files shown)


nexus_workcell_orchestrator/src/workcell_orchestrator.cpp
```diff
@@ -988,8 +978,18 @@ BT::Tree WorkcellOrchestrator::_create_bt(const std::shared_ptr<Context>& ctx)
...
+    RCLCPP_DEBUG(
+      this->get_logger(),
+      "Loading remapped BT [%s] for original task type [%s]",
+      new_task.c_str(),
+      ctx->task.type.c_str()
+    );
+  }
   return this->_bt_factory->createTreeFromFile(this->_bt_store.get_bt(
-        ctx->task.type));
+        new_task));
```
luca-della-vedova, line 992: Note here we could either: 
* Use the new name to only load the behavior tree (what I did)
* Override the task type (what I didn't do)

nexus_common/include/nexus_common/task_remapper.hpp
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Provides task remapping capability
+ */
+class NEXUS_COMMON_EXPORT TaskRemapper
+{
+public:
+  /*
+   * Initialize the remapper with the value of a ROS parameter containing a YAML
+   */
+  TaskRemapper(const std::string& param);
```
Yadunund, line 36: This API is error prone especially if users do not pass a string that can be deserialized into a YAML. ie `YAML::Load()` can throw an exception in the implementation.
I suggest we accept a concrete type here instead like an `std::multimap` where the key  is the remapped process and the values for each key are the process steps in the work order.

This way if the param definition changes in the orchestrators, this API would not have to change.

nexus_common/include/nexus_common/task_remapper.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+  /*
+   * Remaps, if necessary, the input task
+   */
+  std::string remap(const std::string& task) const;
```
Yadunund, line 41: The current implementation returns `task` if no remapping is specified which 1) is an implicit behavior and 2) could give API users a false sense that a remapping was performed. Suggest to have the API return an `std::optional<std::string>` instead of no remapping was specified and handle the optional case in the orchestrator. Yes it adds some duplication in the orchestration logic but the API is clearer and users could potentially save one unnecessary copy of the string here if no remapping is found.

nexus_common/src/task_remapper_test.cpp
```diff
@@ -0,0 +1,48 @@
...
+    R"(
+      pick_and_place: [pick, place]
+      main : ["*"]
+    )";
+  auto remapper = TaskRemapper(param);
+  CHECK(remapper.remap("pick") == "main");
+  CHECK(remapper.remap("place") == "main");
+  CHECK(remapper.remap("other") == "main");
+}
+
```
Yadunund, line 47: Thanks for adding a test! It's great. Could we include one additional testcase 

```
    R"(
      pick_and_place: [pick, *]
    )";
 ```

nexus_workcell_orchestrator/src/workcell_orchestrator.cpp
Yadunund: I prefer what you did which preserves the original request. Users can always look for the remapping if needed from `ctx-> task_remapper`

Yadunund (changes requested):
Minor feedback on the `TaskRemapper` API but this is looking good otherwise!

nexus_common/include/nexus_common/task_remapper.hpp
luca-della-vedova: https://github.com/osrf/nexus/pull/55/commits/b881d1e2046602ce3f909ac9df26412bc5935e94

nexus_common/src/task_remapper_test.cpp
luca-della-vedova: https://github.com/osrf/nexus/pull/55/commits/8c62c69275eebd5361d44ed37f037567347a4598

nexus_common/include/nexus_common/task_remapper.hpp
luca-della-vedova: This PR doesn't introduce any new exception since all uses of this class were doing `YAML::Load` anyway, so they would have thrown on an invalid YAML.
The "proper" way I considered going for was either passing a YAML node as an input (so we won't throw) or keep the string and make the constructor fallible (i.e. a `make` function, return `nullptr` if the YAML failed deserializing). I didn't do it only because I saw that the codebase doesn't seem to use this paradigm but I'm happy to do it.

I feel that if we add a `multimap` and users of the API need to manually parse the YAML, handle exceptions, add its content to a `multimap`, then we read the `multimap`, iterate over it, check for wildcards, populate internal data structures accordingly, it becomes much more complex and clunky. Instead of being a refactor to reduce duplicated code we actually end up with a fair bit more code and complexity.

Yadunund: Let's atleast pass a YAML node and add documentation to explain we expect the dict we expect?

luca-della-vedova: Done in https://github.com/osrf/nexus/pull/55/commits/97fe56cae0ca3b8293422be29e25fa7501a08971. I actually found out in the process that the `TaskParser` was overwriting the task id after all, so I removed that, following feedback in https://github.com/osrf/nexus/pull/55#discussion_r1909596228 that we shouldn't change the task definition itself but only load a different behavior tree

Yadunund (approved):
Thanks for iterating!

PR #69, opened 2025-01-27
Support multiple destinations with actions in nexus_transporter

Yadunund:
This PR
- Updates `nexus_transporter_msgs` to include a `Destination` msg with fields to specify the type of action to perform at the destination.
- Updates the `Itinerary` and `Transporter` classes to support multiple `Destinations`.
- Switches to `SingleThreadedExecutor` (default). This was previously set to `MultiThreadedExecutor` to deal with making a registration client request within a timer callback which leads to a deadlock. This is resolved in the same way the Workcell Orchestrator handles it, ie, relying on async client requests. 
- Move the `mock_transporter` plugin implementation from `nexus_integration_tests` to `nexus_transporter.

As a follow up, I will update rmf_request interfaces introduced in #42 with the Destination definition here and explore a native RMF plugin implementation.


nexus_msgs/nexus_transporter_msgs/msg/Destination.msg
```diff
@@ -0,0 +1,14 @@
...
+string name
+
+# Action constants for the action field.
+int32 ACTION_UNDEFINED=0
+int32 ACTION_TRANSIT=1
+int32 ACTION_PICKUP=2
+int32 ACTION_DROPOFF=3
+
+# The type of action to perform at the destination.
+int32 action
```
aaronchongth, line 11: we probably won't need so many action types in the future, we can use an uint8 for this field

nexus_transporter/src/TransporterNode.hpp
```diff
@@ -102,20 +100,22 @@ class TransporterNode : public rclcpp_lifecycle::LifecycleNode
...
+    rclcpp::Client<RegisterTransporter>::SharedPtr register_client;
+
+    // Asynchronously monitor whether the transport has registered with the
+    // system orchestrator.
+    std::optional<rclcpp::Client<RegisterTransporter>::
+      SharedFutureAndRequestId> ongoing_register;
+
+    rclcpp::TimerBase::SharedPtr register_timer;
+
+
```
aaronchongth, line 113: nit: extra line

nexus_transporter/src/TransporterNode.cpp
```diff
@@ -459,44 +453,72 @@ TransporterNode::TransporterNode(const rclcpp::NodeOptions& options)
...
+      RCLCPP_WARN(node->get_logger(),
+        "Unable to remove pending request during transporter registration.");
+    }
   }
-  auto resp = fut.get();
-  if (!resp->success)
+
+  RCLCPP_INFO(node->get_logger(), "Registering with system orchestrator...");
+  auto register_cb =
+    [this, node](rclcpp::Client<RegisterTransporter>::SharedFuture future)
```
aaronchongth, line 479: we should capture `w_node` instead, and lock it inside `register_cb` since this is an async request and `node` is expected to be released as `_register` returns before we get a response

aaronchongth (changes requested):
This looks good! Left some small comments, let me know what you think

PR #72, opened 2025-02-05
Rename nexus_integration_tests to nexus_demos

aaronchongth:
Just a basic renaming with some minor changes to docs


Yadunund (approved):
Thanks for the migration!

PR #74, opened 2025-02-24
Speed up build by using released bridge binaries

aaronchongth:


aaronchongth:
Closing as there is no elegant way to handle downloading different architectures :grimacing: I'll wait till the migration to `rmw_zenoh` then

PR #79, opened 2025-04-01
Autostarting lifecycle nodes (fixes #77)

### LINKED ISSUE #77: Autostart Lifecycle nodes in launch file using new `autostart` param in launch_ros
Rely on https://github.com/ros2/launch_ros/pull/430 instead of the custom function we have now like https://github.com/osrf/nexus/blob/c4f84218372819dbfc29a061d5679e8cf54318f1/nexus_demos/launch/inter_workcell.launch.py#L37

JRV7903:
I have removed the manual `activate_node` function from `inter_workcell.launch.py` and `nexus_calibration.launch.py` and added the new autostart parameter. I have also removed the `activate_node_service` function from `workcell.launch.py` and integrated the autostart parameter for the `motion_planner_server` in `demo_planner_server.launch.py`.

### DIFF
--- a/nexus_calibration/test/nexus_calibration.launch.py
+++ b/nexus_calibration/test/nexus_calibration.launch.py
@@ -18,84 +18,11 @@ from ament_index_python.packages import get_package_share_directory
 
 import launch_ros
 from launch_ros.actions import LifecycleNode
-from launch_ros.event_handlers import OnStateTransition
 import launch
 from launch.actions import (
-    GroupAction,
     OpaqueFunction,
     ExecuteProcess,
 )
-import lifecycle_msgs
-
-def activate_node(target_node: LifecycleNode, depend_node: LifecycleNode = None):
-
-    configure_event = None
-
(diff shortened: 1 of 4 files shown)


aaronchongth:
Hi @JRV7903, thanks for the PR. Could you GPG sign and signoff on your commits?

I noticed that the integration tests are failing too, where the nodes do not seem to have transitioned. You can run the tests locally according to instructions here, https://github.com/osrf/nexus/blob/main/.github/workflows/nexus_integration_tests.yaml, to verify if your fix works.

aaronchongth (changes requested):
Hi @JRV7903, thanks for the PR. Could you GPG sign and signoff on your commits?

I noticed that the integration tests are failing too, where the nodes do not seem to have transitioned. You can run the tests locally according to instructions here, https://github.com/osrf/nexus/blob/main/.github/workflows/nexus_integration_tests.yaml, to verify if your fix works.

Issue #59, opened 2025-01-13
Exclude logs for certain BT nodes


Yadunund:
By default, we log execution times for [all nodes](https://github.com/osrf/nexus/blob/d16e817af539d0b7d8c67fa88960a166f2cc910d/nexus_common/include/nexus_common/logging.hpp#L90) in a BT. However this leads to terminal spam from nodes like `IsPaused`.

Ideally each BT::Node has an API to specify log level which the `nexus::BtLogging` class respects but this might be more involved.

We could also consider adding an API to `nexus::BtLogging` to exclude logs for certain nodes and add the `IsPaused` to this exclude list.

PR #83, opened 2025-04-08
Allow certain BT nodes to be blocklisted for logging

Yadunund:
Stopgap solution for #59.

This PR updates the `BTLogging` class to accept a list of nodes that that we want to blocklist for logging.

Also updates system and workcell orchestrators to get these node names via a `bt_logging_blocklist` param. The launch file s are updated set this param to blocklist `IsPauseTriggered` which prevents terminal spam when running a process.


aaronchongth (approved):
works great, LGTM

Issue #81, opened 2025-04-08
Possible regression in workcell orchestrator


Yadunund:
When sending goals to the system orchestrator, i'm seeing the WO crash sometimes. Looking at the trace below, it has to do with [aborting the goal handle here](https://github.com/osrf/nexus/blob/f8488f126756fe50881ea86f58fd6febc72d2fdf/nexus_workcell_orchestrator/src/workcell_orchestrator.cpp#L1037).

We might have to take another look at the iterator logic that was changed in #73.

```
[nexus_workcell_orchestrator-9] #20   Object "/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2", at 0xffffffffffffffff, in 
[nexus_workcell_orchestrator-9] #19   Object "/ws_nexus/install/nexus_workcell_orchestrator/lib/nexus_workcell_orchestrator/nexus_workcell_orchestrator", at 0x560a301b66e4, in _start
[nexus_workcell_orchestrator-9] #18   Source "../csu/libc-start.c", line 360, in __libc_start_main_impl [0x7f8de051428a]
[nexus_workcell_orchestrator-9] #17   Source "../sysdeps/nptl/libc_start_call_main.h", line 58, in __libc_start_call_main [0x7f8de05141c9]
[nexus_workcell_orchestrator-9] #16   Object "/ws_nexus/install/nexus_workcell_orchestrator/lib/nexus_workcell_orchestrator/nexus_workcell_orchestrator", at 0x560a301b5fdc, in main
[nexus_workcell_orchestrator-9] #15   Object "/opt/ros/jazzy/lib/librclcpp.so", at 0x7f8de0ae2253, in rclcpp::executors::SingleThreadedExecutor::spin()
[nexus_workcell_orchestrator-9] #14   Object "/opt/ros/jazzy/lib/librclcpp.so", at 0x7f8de0ad1bd0, in rclcpp::Executor::execute_any_executable(rclcpp::AnyExecutable&)
[nexus_workcell_orchestrator-9] #13   Object "/ws_nexus/install/nexus_workcell_orchestrator/lib/libnexus_workcell_orchestrator_plugin.so", at 0x7f8ddcb3bd93, in rclcpp::GenericTimer<nexus::workcell_orchestrator::WorkcellOrchestrator::on_activate(rclcpp_lifecycle::State const&)::{lambda()#1}, (void*)0>::execute_callback(std::shared_ptr<void> const&)
[...]

PR #73, opened 2025-02-18
Parallel workcell task

aaronchongth:
Based off https://github.com/osrf/nexus/pull/71

* Allow workcells to operate multiple BTs at the same time
* Re-arrange BT for place workcell, so it waits for transporter before picking up label. This resolves the issue where a different AMR reaches the workcell before the first task's AMR
* Re-worked the building map, to allow more lanes and mutex groups to facilitate traffic around entry to workcells
* Fix bug where iterators for workcell context still used after destroyed
* Add test

https://github.com/user-attachments/assets/ea9a6261-3655-4151-b05d-0125a37bb973

# Testing

```bash
# launch everything, and start system orchestrator
ros2 launch nexus_demos launch.py headless:=False use_rmf_transporter:=True sim_update_rate:=1000

# start multiple tasks with different ids, and watch them go
ros2 action send_goal /system_orchestrator/execute_order nexus_orchestrator_msgs/action/ExecuteWorkOrder "{order: {work_order_id: '23', work_order: '$(cat config/pick_and_place.json)'}}"

ros2 action send_goal /system_orchestrator/execute_order nexus_orchestrator_msgs/action/ExecuteWorkOrder "{order: {work_order_id: '24', work_order: '$(cat config/pick_and_place.json)'}}"

ros2 action send_goal /system_orchestrator/execute_order nexus_orchestrator_msgs/action/ExecuteWorkOrder "{order: {work_order_id: '25', work_order: '$(cat config/pick_and_place.json)'}}"

ros2 action send_goal /system_orchestrator/execute_order nexus_orchestrator_msgs/action/ExecuteWorkOrder "{order: {work_order_id: '26', work_order: '$(cat config/pick_and_place.json)'}}"
```

PR #84, opened 2025-04-09
Delete failed context before cancelling all other contexts

aaronchongth:
Fixes https://github.com/osrf/nexus/issues/81.

Added a test to check that aborts are happening correctly.

## Cause of bug

The issue ticket was correct that the refactor in https://github.com/osrf/nexus/pull/73 was the cause, as it did not take into account that Failed goals were already aborted upstream, and their context should be removed from the list of contexts, before cancelling all other contexts, to prevent a double cancellation.


nexus_demos/config/workcell_1_bts/invalid_place_on_conveyor.xml
```diff
@@ -0,0 +1,29 @@
+<?xml version='1.0'?>
```
Yadunund, line 1: Could we either reduce the number of steps in the BT (even just have one bad step) or call out which step is invalid via coments?

nexus_demos/config/workcell_1_bts/place_on_conveyor.xml
```diff
@@ -14,7 +14,7 @@
       <execute_trajectory.ExecuteTrajectory name="execute_pickup" trajectory="{pickup_traj}" />
       <gripper.GripperControl name="pickup_product" gripper="workcell_1_mock_gripper" position="0.0" max_effort="1.0" />
 
-      <plan_motion.PlanMotion name="plan_to_dropoff" robot_name="abb_irb910sc" goal="dropoff_pose" scale_speed="0.5" cartesian="true" start_constraints="{joint_constraints}" result="{dropoff_traj}" />
+      <plan_motion.PlanMotion name="plan_to_dropoff" robot_name="abb_irb910sc" goal="invalid_dropoff_pose" scale_speed="0.5" cartesian="true" start_constraints="{joint_constraints}" result="{dropoff_traj}" />
```
Yadunund, line 17: undo?

nexus_demos/test_invalid_place_on_conveyor.py
```diff
@@ -0,0 +1,87 @@
+# Copyright (C) 2022 Johnson & Johnson
```
Yadunund, line 1:
```suggestion
# Copyright (C) 2025 Open Source Robotics Foundation
```

Yadunund (changes requested):
Thanks for hunting down the bug! 
Left a couple of minor comments to address.

nexus_demos/config/workcell_1_bts/place_on_conveyor.xml
aaronchongth: https://github.com/osrf/nexus/pull/84/commits/802222fceb850e42d9c3416664a8b39340474c66

nexus_demos/test_invalid_place_on_conveyor.py
aaronchongth: https://github.com/osrf/nexus/pull/84/commits/7ab4f2572e3c7da6d19f535ecd5baa922ad248b6

nexus_demos/config/workcell_1_bts/invalid_place_on_conveyor.xml
aaronchongth: https://github.com/osrf/nexus/pull/84/commits/7ab4f2572e3c7da6d19f535ecd5baa922ad248b6

Yadunund (approved):
LGTM!

PR #88, opened 2025-04-17
Update default  WorkOrder model

Yadunund:
This PR
- Gets rid of `Item` as a required field in the work order. This is replaced by `MetaData` which is optional in the work order. Users can directly access the yaml contents.
- Gets rid of `name` inside `Step` as it was redundant.
- Adds a getter for `process_parameters` in `Step` to return a YAML node of parameters. Downstream skills should have access to these parameters via `Task.data ` in the [context during configuration](https://github.com/osrf/nexus/blob/eb61acfab2cf44235ff132507a803e968ae92f9b/nexus_capabilities/include/nexus_capabilities/capability.hpp#L59)

### DIFF
--- a/nexus_common/include/nexus_common/models/metadata.hpp
+++ b/nexus_common/include/nexus_common/models/metadata.hpp
@@ -1,0 +1,59 @@
+/*
+ * Copyright (C) 2025 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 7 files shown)


aaronchongth (approved):
Looks good! Just left a question for consideration, but I'll leave it up to you since you have better context on the use-case or non-use-case.

aaronchongth (approved):
LGTM

PR #89, opened 2025-04-23
Get transporter itinerary asynchronously

aaronchongth:
* allows transporters to get itineraries in an asynchronous manner
* will be required for dry-run bidding for the rmf transporter down the road
* add mutex for itineraries in transporter node
* pass copies for itineraries, to prevent dangling references since we cannot guarantee its lifetime in the transporter node
* re-instated uncrustify, some minor linting


nexus_transporter/include/nexus_transporter/Transporter.hpp
```diff
@@ -58,18 +60,22 @@ class Transporter
...
   /// \param[in] job_id
   /// An id for this request.
   ///
   /// \param[in] destinations
   /// A list of destinations.
-  /// \return A nullopt is returned if the destinations is not valid.
+  ///
+  /// \param[in] completed_cb
+  /// A callback to execute when an itinerary query has been completed.
   // TODO(YV): Consider creating a separate class for destination
```
Yadunund, line 74: Unrelated but we can remove this TODO.

Yadunund (approved):
LGTM!

nexus_transporter/include/nexus_transporter/Transporter.hpp
aaronchongth: https://github.com/osrf/nexus/pull/89/commits/67b242227cba288ce15199f38b1584d82c234558

PR #91, opened 2025-05-24
Allow configuring RMF simulation args

Yadunund:
So that these args can be overwritten by other launch files.

### DIFF
--- a/nexus_demos/launch/rmf_transporter.launch.xml
+++ b/nexus_demos/launch/rmf_transporter.launch.xml
@@ -75,11 +75,9 @@
   <!-- Simulator launch -->
   <let name="gz_headless" if="$(var headless)" value="-s"/>
   <let name="gz_headless" unless="$(var headless)" value="" />
-  <let name="world_path" value="$(var gz_world_file)" />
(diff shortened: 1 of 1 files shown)


nexus_demos/launch/rmf_transporter.launch.xml
```diff
@@ -69,11 +75,11 @@
   <!-- Simulator launch -->
   <let name="gz_headless" if="$(var headless)" value="-s"/>
   <let name="gz_headless" unless="$(var headless)" value="" />
-  <let name="world_path" value="$(find-pkg-share nexus_demos)/config/rmf/maps/$(var building_map_name)/$(var building_map_name).world" />
-  <let name="model_path" value="$(find-pkg-share nexus_demos)/config/rmf/maps/$(var building_map_name)/models" />
+  <let name="world_path" value="$(var gz_world_file)" />
+  <let name="model_path" value="$(var gz_model_path)" />
```
aaronchongth, line 79: We can probably just use `gz_world_file` and `gz_model_path` directly then

aaronchongth (approved):
LGTM

PR #93, opened 2025-06-04
Cancel callbacks for BT nodes in nexus_common

Yadunund:
A minor improvement to allow users to override a cancellation callback that will be called when a BT node in `nexus_common` is halted.


aaronchongth (approved):
Thanks, this is helpful, LGTM

PR #95, opened 2025-06-04
Do not overwrite task id in DispatchRequest

Yadunund:
This PR reverts the logic introduced in #94 where the `DispatchReqeust` skill overwrites the `task_id` for the workcell. 

Turns out overwriting the `task_id` in the workcell prevents it from receiving [signals](https://github.com/osrf/nexus/blob/main/nexus_system_orchestrator/src/send_signal.cpp#L47-L48) from the system orchestrator 😅 

I removed the  `if (msg->request_guid != rmf_workcell_state_->task_id)` check in the `rmf_task_canceller` node as this was just an extra safety check. I've tested functionality locally and things work as expected.


aaronchongth (approved):
Ah right, the signals are specific about workcell tasks too. Thanks for fixing that LGTM

PR #98, opened 2025-06-09
Return error code on colcon test failure

### LINKED ISSUE #97: CI doesn't detect test failures anymore
Surfaced in https://github.com/osrf/nexus/pull/92#issuecomment-2954784110. Since https://github.com/osrf/nexus/pull/87, specifically moving from `python3 -m unittest` to `colcon test`, integration tests don't actually detect failure in CI.

The reason for that is that by default `colcon test` only returns a non-zero code if it _fails in running a test_, not if there are test failures. If we wanted it to return non zero code on test failure there is a flag for it as noted in the [docs](https://colcon.readthedocs.io/en/released/reference/verb/test.html) `--return-code-on-test-failure`.
By contrast, `unittest` returns non zero code on test failures.

In alternative, we could revert to using the previous command to run the integration tests, or use the ROS ci action.

luca-della-vedova:
Closes #97, just configure colcon to return a non-zero code on test failure.

### DIFF
--- a/.github/workflows/nexus_integration_tests.yaml
+++ b/.github/workflows/nexus_integration_tests.yaml
@@ -43,4 +43,4 @@ jobs:
         timeout_minutes: 20
         max_attempts: 3
         retry_on: error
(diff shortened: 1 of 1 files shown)


aaronchongth (approved):
LGTM, thanks! I'll merge `main` into https://github.com/osrf/nexus/pull/92 after this is merged, and we should expect CI to fail

PR #107, opened 2025-07-16
Build basic RMF image, to be optionally used for CI

aaronchongth:
With the ongoing development and integration of the RMF transporter #104 and #106 , we will need newer RMF features unreleased into jazzy. This workflow will allow us to manually trigger image builds that will have RMF source built


luca-della-vedova (approved):

aaronchongth:
I'll revisit this, and perhaps remove this if we manage to get a similar workflow upstream on open-rmf/rmf

PR #110, opened 2025-07-25
RMF nexus transporter use RMF bidding

aaronchongth:
Before, we would naively check if the navigation graph contains the workcell name as a waypoint. With this PR, we will do that as a first check, before running the proper RMF bidding.

Notable changes:
* added an internal node to handle all RMF related interactions, to not be blocked by transporter node's services or actions
* transporters still use single threaded executors, with the RMF transporter using an internal thread for spinning

https://github.com/user-attachments/assets/7ff18d29-dba4-47b9-8b84-6ab2791ef918

### DIFF
--- a/nexus_transporter/src/transporters/rmf_transporter.cpp
+++ b/nexus_transporter/src/transporters/rmf_transporter.cpp
@@ -399,11 +399,9 @@ public:
 
     _spin_thread = std::make_shared<std::thread>([&]()
       {
-        while (rclcpp::ok())
-        {
-          rclcpp::spin_some(_internal_node);
-          std::this_thread::sleep_for(std::chrono::milliseconds{100});
-        }
+        rclcpp::experimental::executors::EventsExecutor executor;
+        executor.add_node(_internal_node);
+        executor.spin();
       });
 
     _auctioneer = rmf_task_ros2::bidding::Auctioneer::make(


nexus_transporter/src/transporters/rmf_transporter.cpp
```diff
@@ -575,8 +745,7 @@ class RmfTransporter : public Transporter
       return;
     }
 
-    // Naively just checks if the waypoints exist on the building map.
-    // TODO(ac): use auctioneer for proper bidding
+    // Checks through building map first
```
luca-della-vedova, line 748: Can we get rid of this altogether now?

nexus_transporter/src/transporters/rmf_transporter.cpp
```diff
@@ -457,57 +622,62 @@ class RmfTransporter : public Transporter
...
-                    << std::endl;
-          return;
-        }
-        auto it = _rmf_task_id_to_ongoing_itinerary.find(msg->request_guid);
-        if (it == _rmf_task_id_to_ongoing_itinerary.end())
+    _dispenser_request_sub =
+      _internal_node->create_subscription<DispenserRequest>(
+        "/dispenser_requests",
+        100,
+        [&](DispenserRequest::SharedPtr msg)
```
luca-della-vedova, line 629: Cries a bit in merge conflicts for #101, I wonder if we make the diff a bit smaller by only moving the relevant task pub/subs into the internal node? I suppose that would be the api requests / responses?

luca-della-vedova:
Also we should (sigh) update the transporter README to reflect changes in the behavior

nexus_transporter/src/transporters/rmf_transporter.cpp
aaronchongth: I was planning on keeping it, as it's a rather cheap lookup as a first check, before going through the bidding process

nexus_transporter/src/transporters/rmf_transporter.cpp
aaronchongth: We can wait for #101 to be merged first, and I'll handle the merge conflicts later then. I think for now a quick look over the implementation is good enough.

I think it's better to isolate all the interactions to the internal node, so that we are sure that there will not be any race conditions or deadlocks if we use mutexes.

nexus_transporter/src/transporters/rmf_transporter.cpp
```diff
@@ -251,18 +393,42 @@ class RmfTransporter : public Transporter
...
+    // being blocked by the transporter node's actions and services
+    _internal_node = rclcpp::Node::make_shared("rmf_transporter_internal_node");
+
+    _spin_thread = std::make_shared<std::thread>([&]()
+      {
+        while (rclcpp::ok())
+        {
+          rclcpp::spin_some(_internal_node);
+          std::this_thread::sleep_for(std::chrono::milliseconds{100});
+        }
```
Yadunund, line 406: I think we can avoid this `spin_some` pattern with just `spin` since this is running in a separate thread. `spin()` should exit when a Ctrl+C signal is registered.

I kinda want to try the `[EventsExecutor](https://github.com/ros2/rclcpp/blob/b7e4aad091c6b334b9ad17a8f83007ea9901cc3b/rclcpp/include/rclcpp/experimental/executors/events_executor/events_executor.hpp#L73)` here for better performance. It's still "experimental" but I think it would be perfect for this sort of internal node. Shall we give this a shot?
```
 rclcpp::experimental::executors::EventsExecutor executor;
 executor.add_node(node);
 executor.spin();
```

Yadunund (changes requested):
This looks great. Left a comment about avoiding the `spin_some` pattern but other than that, LGTM.

nexus_transporter/src/transporters/rmf_transporter.cpp
aaronchongth: great idea, swapped it over in https://github.com/osrf/nexus/pull/110/commits/4415f9edee6985d02b33e7d0757eb99b583ee5cb

Issue #113, opened 2025-08-07
Error building nexus_robot_controller when following README instructions


b-corry:
Following the README instructions through https://github.com/osrf/nexus?tab=readme-ov-file#build-the-nexus-workspace

I am seeing an error with the `nexus_robot_controller` package when running the build step:
```bash
colcon build --cmake-args -DCMAKE_BUILD_TYPE=Release
```

Output:
```bash
-- Configuring incomplete, errors occurred!
--- stderr: nexus_robot_controller                         
CMake Error at CMakeLists.txt:25 (target_link_libraries):
  Cannot specify link libraries for target "{robot_controller_server_core"
  which is not built by this project.


---
Failed   <<< nexus_robot_controller [4.72s, exited with code 1]

Summary: 4 packages finished [7.72s]
  1 package failed: nexus_robot_controller
  1 package had stderr output: nexus_robot_controller

```

PR #114, opened 2025-08-07
Fixed typo in CMakeLists.txt causing build failure

b-corry:
Fixed bug causing https://github.com/osrf/nexus/issues/113#issue-3301828503

Tested:
1. Followed the README setup instructions through https://github.com/osrf/nexus?tab=readme-ov-file#build-the-nexus-workspace
2. Confirmed no failed packages from the step: 
```bash
colcon build --cmake-args -DCMAKE_BUILD_TYPE=Release
```

### DIFF
--- a/nexus_robot_controller/CMakeLists.txt
+++ b/nexus_robot_controller/CMakeLists.txt
@@ -22,7 +22,7 @@ endforeach()
 
 # Library
 add_library(${library_name} src/robot_controller_server.cpp)
-target_link_libraries({${library_name} PUBLIC
+target_link_libraries(${library_name} PUBLIC
   controller_manager::controller_manager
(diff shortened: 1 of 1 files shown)


luca-della-vedova (commented):
Nice find! Can you add the `nexus_robot_controller` packages to our CI [here](https://github.com/osrf/nexus/blob/17ca92ba4f7933ef39a3ebb40e0482a1c26fe2ed/.github/workflows/nexus_unit_tests.yaml#L39-L42) so that we make sure this doesn't happen in the future?

luca-della-vedova (approved):
Imma go and ignore flaky integration tests since there is no way this PR broke them, thanks for the fix!

PR #115, opened 2025-08-11
Increase setup timeout for RMF integration test

luca-della-vedova:
We have been seeing increased flakiness on RMF integration tests for some time.
I looked at the logs / tried to reproduce locally and I'm fairly confident it's due to the fact that our worlds use very large models ([depot](https://app.gazebosim.org/OpenRobotics/fuel/models/Depot) itself is almost 80MB!) and we expect the bringup, which includes downloading all the models, in 60 seconds.

I tried on a very high power machine and downloading models seems to add a good chunk of time to the integration test, I also noticed that gazebo adds some printouts when it finished downloading models and our CI times out and fails before the printout is logged.

Trying to double the timeout to see if it helps.


luca-della-vedova:
Fixed by #121 instead

PR #121, opened 2025-08-19
Remove ccache from CI

luca-della-vedova:
Do you know why CI has been flaky for ages?

Because specifying `--cmake-args` after the mixin overwrote the `release` mixin and we were building in debug mode.

There.

### DIFF
--- a/.github/workflows/nexus_integration_tests.yaml
+++ b/.github/workflows/nexus_integration_tests.yaml
@@ -30,13 +30,12 @@ jobs:
     - name: rosdep
       run: |
         apt update
-        apt install -y ccache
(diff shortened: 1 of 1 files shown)


aaronchongth (approved):
Thanks for the fix! LGTM

Yadunund:
Why remove ccahce? Wouldn't this lead to slower builds?

aaronchongth:
Yeah looking at the build-times, I think we weren't using cccahe together with the `action/cache` properly.

From some quick searches, it looks like the steps required to restore the cache is a little more convoluted than what we were doing. I'd be happy to take a crack at it when I'm done with the rmw migration, but hopefully by then the compilation times would have sped up enough to not be an issue already

PR #123, opened 2025-08-21
Move `rmf_canceller` logic into RMF transporter

Yadunund:
Started working on this but later realized the logic in this entire script needs to be moved into the `rmf_transporter` plugin. Leaving this open as a reference for ingestor logic.


aaronchongth:
Some changes here also refactors the expected mappings between itinerary IDs and RMF task IDs, in order to check which RMF task IDs are related to which work order, and if those work orders have been cancelled

Yadunund (commented):
Tested and works great!

PR #125, opened 2025-08-28
Parameterize mock transporter world frame

aaronchongth:
Allows users to remap the parent frame name of the mock transporter when publishing the feedback. By default use the `world` frame name. The transporter node will check if the feedback populated the frame name, otherwise will default to `world` as well.

### DIFF
--- a/nexus_transporter/src/TransporterNode.cpp
+++ b/nexus_transporter/src/TransporterNode.cpp
@@ -399,14 +399,9 @@ auto TransporterNode::on_configure(const State& /*previous_state*/)
               geometry_msgs::msg::TransformStamped tf_msg;
               tf_msg.header.stamp =
               node ? node->get_clock()->now() : rclcpp::Clock().now();
-              if (state.location.header.frame_id.empty())
-              {
-                tf_msg.header.frame_id = "world";
-              }
-              else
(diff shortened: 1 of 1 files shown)


nexus_transporter/src/TransporterNode.cpp
```diff
@@ -399,7 +399,14 @@ auto TransporterNode::on_configure(const State& /*previous_state*/)
...
               node ? node->get_clock()->now() : rclcpp::Clock().now();
-              tf_msg.header.frame_id = "world";
+              if (state.location.header.frame_id.empty())
+              {
+                tf_msg.header.frame_id = "world";
+              }
+              else
+              {
+                tf_msg.header.frame_id = state.location.header.frame_id;
+              }
```
Yadunund, line 409: Nit: Can rely on ternary expression for assignment.

Yadunund (approved):
LGTM

nexus_transporter/src/TransporterNode.cpp
aaronchongth: https://github.com/osrf/nexus/pull/125/commits/08fa8bfab98f68b63753d663d3bd77e5f813d700

PR #129, opened 2025-10-08
Remove deprecated ign calls

luca-della-vedova:
I was just trying to test `nexus` with Ionic and errors popped up because we are using deprecated `ign` APIs, just change them to `gz` instead


aaronchongth (approved):
LGTM!

PR #138, opened 2025-12-17
revert loop change

aaronchongth:
## What's new

Fixes a bug introduced when addressing review feedback in https://github.com/osrf/nexus/pull/137

* loop cannot be combined as we need to gather all assignments, before adding to all assignments
* not using range based loop (2nd one), because it is used to populate index

### GenAI Use
We follow [OSRA's policy on GenAI tools](https://github.com/openrobotics/osrf-policies-and-procedures/blob/main/OSRF%20Policy%20on%20the%20Use%20of%20Generative%20Tools%20(%E2%80%9CGenerative%20AI%E2%80%9D)%20in%20Contributions.md)

- [ ] I used a GenAI tool in this PR.
- [x] I did not use GenAI

Generated-by:

### DIFF
--- a/nexus_system_orchestrator/src/add_task_assignments_to_metadata.cpp
+++ b/nexus_system_orchestrator/src/add_task_assignments_to_metadata.cpp
@@ -26,26 +26,31 @@ BT::NodeStatus AddTaskAssignmentsToMetadata::tick()
 {
   auto tasks = this->_ctx->get_tasks();
 
-  YAML::Node assignment_result;
-  assignment_result["workcell_assignments"] = YAML::Load("[]");
-  assignment_result["current_index"] = 0;
-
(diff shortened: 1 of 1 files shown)


luca-della-vedova (approved):
Ahhh yes now I see it, sorry for the misguided review :facepalm: btw I was referring to the first loop for the range based iteration, I agree that the second cannot be changed because it needs the index

aaronchongth:
ah ok. I'm a little old fashioned, I kinda like that it's clearer if we traverse the vector in the same manner on both loops, index based on both, rather than range based then index based. (despite the guarantees made by std lib :sweat_smile:), especially since we want to keep track of indices

it should not matter in terms of performance, since the `.at()` call should return a const ref anyway

thanks for the quick review!

